PR #67, opened 2021-02-18
Docker CP instead of rsync builds in project/.gitlab-ci-local

firecow:


src/job.ts
```diff
@@ -248,24 +251,20 @@ export class Job {
...
-            }
-            preCmd += `"\n`;
-            await Utils.spawn(preCmd, this.cwd);
-            const endTime = process.hrtime(time);
-            process.stdout.write(`${this.getJobNameString()} ${magentaBright(`rsync to build folder`)} in ${magenta(prettyHrtime(endTime))}\n`);
+            time = process.hrtime();
+            process.stdout.write(`${jobNameStr} ${magentaBright(`pulling`)} ${this.image}\n`);
+            let pullCmd = ``;
+            pullCmd += `docker image ls --format '{{.Repository}}:{{.Tag}}' | grep -E '^${this.image}$'\n`
+            pullCmd += `if [ "$?" -ne 0 ]; then\n`
```
ZobairQ, line 258: wouldn't this look better as one expression?

firecow: Sure, if we don't care about line length 😃
And I don't so will change

FnuGk (approved):
LGTM

src/job.ts
firecow: Wait, we cannot put all this on one line it will be way too long

firecow:
```
            pullCmd += `docker image ls --format '{{.Repository}}:{{.Tag}}' | grep -E '^${this.image}$'\n`
            pullCmd += `if [ "$?" -ne 0 ]; then\n`
            pullCmd += `\techo "Pulling ${this.image}"\n`
            pullCmd += `\tdocker pull ${this.image}\n`
            pullCmd += `fi\n`
```

ZobairQ (approved):
LGTM

PR #86, opened 2021-03-06
Extrapolate predefines from git remote

firecow:
- Made predefined values extrapolate values from git remote.
- Added JobOptions interface instead of passing 1million params to Job class.
- Introduced assert function. Made throw new logic cleaner


moberghammer (approved):
LGTM

PR #95, opened 2021-03-15
Implement include:template support

firecow:


src/parser.ts
```diff
@@ -390,7 +395,11 @@ export class Parser {
                 const localDoc = await Parser.loadYaml(`${cwd}/${value.local}`);
                 includeDatas = includeDatas.concat(await Parser.prepareIncludes(localDoc, cwd, gitRemote, tabCompletionPhase));
             } else if (value["file"]) {
-                const fileDoc = await Parser.loadYaml(`${cwd}/.gitlab-ci-local/includes/${value["project"]}/${value["ref"] || "master"}/${value["file"]}`);
+                const fileDoc = await Parser.loadYaml(`${cwd}/.gitlab-ci-local/includes/${gitRemote.domain}/${value["project"]}/${value["ref"] || "master"}/${value["file"]}`);
```
moberghammer, line 398: I think you gonna change "master" as default since this is being abandoned for "main". 
I think this is configurabale

src/parser.ts
```diff
@@ -344,8 +344,8 @@ export class Parser {
 
     static async downloadIncludeFile(cwd: string, project: string, ref: string, file: string, gitRemoteDomain: string): Promise<void> {
         const time = process.hrtime();
-        fs.ensureDirSync(`${cwd}/.gitlab-ci-local/includes/${project}/${ref}/`);
-        await Utils.spawn(`git archive --remote=git@${gitRemoteDomain}:${project}.git ${ref} ${file} | tar -xC .gitlab-ci-local/includes/${project}/${ref}/`, cwd);
+        fs.ensureDirSync(`${cwd}/.gitlab-ci-local/includes/${gitRemoteDomain}/${project}/${ref}/`);
```
moberghammer, line 347: There can be more then one remote in a git checkout which do you use?

src/parser.ts
ZobairQ: You can configure it from github under settings

firecow: Gitlabs repo use master, so it doesn't matter... But what does that have to do with this line of code ?

src/parser.ts
firecow: The one that i've always used. What does this have to do with this Pull Request ?

ZobairQ (approved):
LGTM

src/parser.ts
moberghammer: you default to master

firecow: Yes, i've always done that. Got nothing to do with this particular feature @moberghammer

moberghammer (approved):
LGTM

PR #112, opened 2021-03-27
Fix bug with extends in multi levels. extends tag is deep-merge, not deep-extend

firecow:
…merge, not deep-extends like the rest of the YAML


bcouetil (approved):
LGTM

PR #115, opened 2021-04-01
Implemented include: { remote: urlString }

firecow:


src/utils.ts
```diff
@@ -34,6 +34,10 @@ export class Utils {
         });
     }
 
+    static fsUrl(url: string): string {
```
moberghammer, line 37: Could a remote be git+ssh?

firecow: No, only http

moberghammer (approved):
Lgtm

PR #121, opened 2021-04-08
Image artifacts in .gitlab-ci-local/artifacts, and cache in /tmp/

bcouetil:


bcouetil:
@firecow how can I make test to go find artifacts dynamically in a folder named after <pipelineIid> ?

bcouetil:
It's OK for now I'm working on a way to refacto tests.

bcouetil:
@firecow help, I can't figure out why my test does not work :sos:

bcouetil:
Both issues implementations works, except tests, due to my low competence in TS/JS :sweat_smile:

src/job.ts
```diff
@@ -338,6 +343,13 @@ export class Job {
                 dockerCmd += `-e ${key}="${String(value).trim()}" `
             }
 
+            if (this.cache && this.cache.key && this.cache.paths) {
+                this.cache.paths.forEach((path) => {
+                    process.stdout.write(chalk`${jobNameStr} {magentaBright mounting cache} for path ${path}\n`);
+                    dockerCmd += `-v \${PWD}/.gitlab-ci-local/cache/${this.cache.key}/${path}:/builds/${path} `
```
firecow, line 349: This will leave root owned files all over peoples project folders.
Using `/tmp/` would be a better place for these to live.

bcouetil:
I managed to sort out my tests 🍾

src/job.ts
bcouetil: By "all over" you mean in .gitlab-ci-local/, the working folder of our tool ?

On my ubuntu, files are not root owned :

```
> ls -lart .gitlab-ci-local/cache/maven/.cache
total 0
-rwxrwxr-- 1 bcouetil plugdev 0 avril  8 16:42 file1
drwxrwxr-- 1 bcouetil plugdev 0 avril  8 16:42 .
drwxrwxr-- 1 bcouetil plugdev 0 avril  8 17:05 ..
```
Can you check on your side ?

firecow: Yes, I shall try it out...

firecow: image
Yup, it will leave root owned files. @bcouetil

bcouetil: OK, it's not ideal, but I guess we have no choice (I'll investigate a bit)

firecow: Its either `/tmp/` or using `docker cp`
`docker cp` magicly makes sure permissions and user/group's are correct. @bcouetil

bcouetil: I made a modification to have files removable by current user, please have a look @firecow 

`docker cp` is great, but Maven produces thousand of files, and would be significantly slower.

bcouetil: I'm changing my mind.

there is a `docker cp`of the repo at the beginning of each job, and this copies the cache also.

I'm moving it to `/tmp/` as you said, to save time (I'm actually loosing 10s at the start of each job).

firecow: If you are copying cache in and out using docker cp the tmp folder isn't needed.

bcouetil: The `docker cp` I was mentioning is the one that populate `/build/`, not related to cache.

I still prefer not to copy cache while the `tmp` alternative is satisfying, because this would be 10s lost on each job in Maven context.

bcouetil: Is it OK for you ? Then I guess we could merge :smile:

src/job.ts
```diff
@@ -377,6 +396,7 @@ export class Job {
         cp.stdin.write(`set -eo pipefail\n`);
 
         if (this.imageName) {
+            cp.stdin.write(`umask 000\n`); // this allows cache files to be managed on host
```
firecow, line 399: Can you explain to me the implications of setting umask to 000 ?
And is there some way to test how remote runners set the umask to?

firecow: Nvm... Found out...
image
Is there something preventing us from using the default 0002 umask. Like remote runners do ?

firecow: Scratch that, its the image that determines the umask, so we definently can't override like this.
image

firecow:
```yaml
job:
  image: alpine
  script:
    - umask

test:
  image: debian
  script:
    - umask
```

bcouetil: I set umask to be able to delete cache without being root.

Now that files are in /tmp/ that is not that important, so I'll delete this line.

firecow:
@bcouetil I'll give it one last review, then we shall get this master piece into ze master

firecow (approved):
This is a good way of handling docker run cache and artifacts, only a proper full blown artifacts/dependency graph is sorted out.

firecow:
@bcouetil Let me know, when you done.. Then i shall merge it.

src/job.ts
bcouetil: Done

bcouetil:
Change requested all done, final round of local test too ; merge when you want :smile:

firecow:
@bcouetil I'm merging now. Lets test via master build for a while, and then I'll release a version.

PR #124, opened 2021-04-09
Inject ssh agent decorator

firecow:


bcouetil (approved):
Very interesting ! Can be handy, even if I have no use case as of today.

I suppose that it is to mimic the real runner which has already a buit-in ssh agent ?

moberghammer (approved):
Lgtm

PR #129, opened 2021-04-11
Mock write streams, so cases.test.ts becomes simpler

firecow:
Centralize stdout/stderr outputs in a class, that can be mocked.


ZobairQ (approved):
LGTM

moberghammer (commented):
LGTM

PR #132, opened 2021-04-13
Fix Docker CP bug

firecow:


bcouetil:
Thanks, I'm trying that and report ASAP

bcouetil:
Still keeping that in mind.

Note : I've spoken of this tool to my 500+ devs IT company, hence the bump in stars :blush:

bcouetil:
I run into the bug at least once a day. Next week I'll switch to the fix, and if it does not happen anymore in a few days, I'll consider this fixed !

firecow:
@bcouetil Yeah, try it out for a week.
I must admit, that this "fix" is pure speculation on my behalf...
I find it extremely weird that the docker create command would exit, before the containers is actually ready. 
But then again, who knows :smile:

bcouetil:
Maybe this can help the case : 

Screenshot from 2021-04-17 22-23-29

There can be no containerId in output. And that would explain the error mentioned in the issue : `must specify at least one container source`

Maybe launching multiple containers at the same time make some output with nothing and others with 2 containers ? Should we queue containers creation ?

firecow:
Aaaa.... Nice...
I'm setting up a 1000 docker create example to be able to reproduce this more consistently @bcouetil 😃

firecow:
image

image

I can reproduce every single time.
It looks like a bug with docker create if you ask me @bcouetil

firecow:
image

This bash script doesn't leave any "empty" container id's

@bcouetil Its a nodejs (Utils.spawn) problem, somehow...

bcouetil:
You could just spawn again on empty output, easy workaround ?

firecow:
@bcouetil My current setTimeout(func, 10) "hack" works with 1000x docker create.
Apparently the stdout and stderr's can be slashed by nodejs somehow, if enough child processes are spawned.

firecow:
@bcouetil I'm merging this into master, if you approve.

bcouetil (approved):
Wonderful ! Approved

PR #137, opened 2021-04-17
Colored End Report

firecow:
image

image

@ZobairQ @bcouetil Shall we run with this for now ?


bcouetil:
This is concise and looks great. LGTM.

My dream is that, one day, we have something like this :

```
✅ job-a --> ✅ job-b --> ❌ job-d
             ✅ job-c
```

But that is only cherry on the cake, we have more important things to tackle.

bcouetil (approved):

firecow:
> This is concise and looks great. LGTM.
> 
> My dream is that, one day, we have something like this :
> 
> ```
> ✅ job-a --> ✅ job-b --> ❌ job-d
>              ✅ job-c
> ```
> 
> But that is only cherry on the cake, we have more important things to tackle.

Ya, that looks awesome, but for now, I'm just trying to make it more noticeable, when things WARN and or FAILS

PR #140, opened 2021-04-17
Made it possible to call docker create --add-host via an --extra-host cli array option

firecow:


firecow:
image

image

I can reproduce every single time now. Nice..
It looks a bug with `docker create` if you ask me @bcouetil

firecow:
@bcouetil This one as well... Then I shall release new version...

bcouetil (approved):
Tested with multiple arguments, working good, and with switched host/ip, error message clear.

Approved ! thanks !

PR #144, opened 2021-04-21
Remove the pipeline id from the .gitlab-ci-local/artifacts folder

firecow:


bcouetil:
Did not see code yet, but we should clean the folder at pipeline start, to not have old artifacts pretending to be for the new pipeline.

firecow:
@bcouetil Only on pipelines and "needs" pipelines right ?

bcouetil:
yes I think you are right, to have best behavior.

firecow:
@bcouetil Done

bcouetil (approved):
LGTM

PR #146, opened 2021-04-21
Only copy non-ignored files to containers

firecow:


moberghammer (approved):
LGTM

firecow:
@moberghammer Can you try it out on one of your big projects. And paste the rsync timings?

firecow:
@bcouetil and @moberghammer I'm awaiting rsync measurements from you guys, to make sure we don't impose unreasonable cost to everyone using the tool.

bcouetil:
Yes I don't forget, it's just that I have been busy lately. Please don't rush this, it's supposed to gain performance, it should not be otherwise 😊

firecow:
@bcouetil No rush, adding `WIP` to pull request

bcouetil:
I tested it.

You say untracked, but I guess you mean ignored, because new files get into containers (I was afraid of that, being forced to commit new files, but it is OK).

I loose multiple seconds in the process. But I think we can do better : If I'm not mistaken, you rsync the exact same data to n folders (for n jobs), then copy this to each one.

Either we should :
1) rsync once, copy each job
1) rsync each job, simply mount folders

 I think I prefer 1., there may be root files in 2.
 
 What do you think ?

firecow:
Copy takes longer than rsync doesn't it?

firecow:
Volume mounting will leave root owned files yeah....

firecow:
Oh, you mean docker copy, not fs copy

firecow:
Well, we gonna need shell-executor eventually, so having a build folder for each job, will be a requirement soon...

bcouetil:
> We gonna need shell-executor build folder isolation eventually, so having a build folder per job, will be a requirement soon...

I always forget that, sorry.

So I suggest this algo :

```
rsync root files to .gitlab-ci-local/builds/.docker/  <-- we are safe because a job cannot start with "."
for each stage
  in parallel, for each job of the stage
    if docker job
      docker cp from .gitlab-ci-local/builds/.docker/
    else
      rsync root files to .gitlab-ci-local/builds/job-name
    fi
    [...]
  end
end
```

This way, there is only one rsync for docker jobs and one rsync for each shell-job. Most of the time, with only docker jobs, there is only one set of duplicated files.

firecow:
Elegant.... :smile: On it

firecow:
@bcouetil It's pushed

bcouetil:
Tested it again, this works great 💪

But I had to do multiple ctrl+c in the middle of a stage with 7 jobs, and each time I have this error :

Screenshot from 2021-04-28 10-01-31

Wouldn't it be better to ignore docker rm errors ?

```
    public async removeContainer() {
        if (this.containerId) {
            try {
                await Utils.spawn(`docker rm -f ${this.containerId}`);
            } catch (e) {
                // false alarm, deletion is already in progress
            }
        }
    }
```

firecow:
> Tested it again, this works great 
> 
> But I had to do multiple ctrl+c in the middle of a stage with 7 jobs, and each time I have this error :
> 
> Screenshot from 2021-04-28 10-01-31
> 
> Wouldn't it be better to ignore docker rm errors ?
> 
> ```
>     public async removeContainer() {
>         if (this.containerId) {
>             try {
>                 await Utils.spawn(`docker rm -f ${this.containerId}`);
>             } catch (e) {
>                 // false alarm, deletion is already in progress
>             }
>         }
>     }
> ```

Yeah, i'll make an issue on that.
@bcouetil Shall we merge ?

bcouetil (approved):
Yes, LGTM

PR #147, opened 2021-04-23
Make it possible to cache absolute paths

firecow:


firecow:
This isn't working on remote runners, nor should it work locally.

PR #149, opened 2021-04-24
Prevent hang forever

firecow:


src/job.ts
```diff
@@ -437,36 +443,35 @@ export class Job {
...
-            shell: "bash",
-            stdio: ["pipe", "pipe", "pipe"],
-            cwd: this.cwd,
-        });
-
-        cp.stdin.write("set -eo pipefail\n");
+        let cmd = "set -eo pipefail\n";
+        cmd += "exec 0< /dev/null\n";
 
         if (this.imageName) {
```
moberghammer, line 449: Maybe introduce imageJob instead of all the if this.imageName

firecow: You talking about a JobBase.ts, ShellJob.ts and DockerJob.ts pattern? @moberghammer

src/job.ts
```diff
@@ -312,6 +312,21 @@ export class Job {
         return `--env SSH_AUTH_SOCK=${process.env.SSH_AUTH_SOCK} -v ${process.env.SSH_AUTH_SOCK}:${process.env.SSH_AUTH_SOCK}`;
     }
 
+    private generateScriptsCmd(scripts: string[]) {
```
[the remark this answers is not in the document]
firecow: It has been renamed.

src/job.ts
moberghammer: Yer something like that so you only need a single if

firecow: Yay, its a lot of work for this. But its a good idea. Won't do it in this PR.

moberghammer (approved):
LGTM

Issue #110, opened 2021-03-27
Pipeline should stop when a needed job fails, when option --needs is used


bcouetil:
# Love 

The more I learn and use this tool the more I love it 🥰

# Request

## Description

Pipeline should stop when a needed job fails using `--needs`.

## Symptom

```
needed:
  stage: .pre
  image: alpine
  script:
    - exit 1

needs:
  image: alpine
  needs:
    - needed
  script:
    - echo "this should not be played"
```

Both jobs play :

```
yml files processed in 39 ms
needed    starting curlimages/curl:7.69.1 (.pre)
needed    $ exit 1
needed    finished in 5.17 s exited with code 1 
needs     starting curlimages/curl:7.69.1 (test)
needs     $ echo "this should not be played"
needs     > this should not be played
needs     finished in 5.17 s

successful needs
failure needed
```

# Workaround

## Description

Ctrl+C

## Impact

- ✅ low : needs actions once before/after a debugging session
- medium : fixed using another top level .gitlab-ci.yml
- high : needs deep refactoring, everywhere in templates or everytime or both

firecow:
@bcouetil This only happens when calling `gitlab-ci-local needs --needs` right ?

bcouetil:
Yes, I've updated the description

PR #151, opened 2021-04-24
Handle jobs execution order properly

firecow:


firecow (commented):
Just review'ed everything here on github. LGTM

src/commander.ts
```diff
@@ -3,126 +3,56 @@ import {Job} from "./job";
...
-            }
-        }
-
-        for (const job of jobs) {
-            await job.start(privileged);
+            jobPoolMap = new Map<string, Job>(jobs);
+            jobArgs.forEach(jobName => {
+                const job = Utils.getJobByName(jobs, jobName);
+                jobPoolMap.set(jobName, job);
+                potentialStarters = potentialStarters.concat(JobExecutor.getPastToWaitFor(jobs, stages, job));
```
moberghammer, line 38: Move if needs to this line only to depublicate

firecow: Job pool map is also special

moberghammer (approved):
Huge PR but looks okay as far as i can tell.

bcouetil (approved):
Tested  #110, it's working.

Too big for me to understand the code. Maybe you could explain the change in philosophy, for us to detect flaws in principle ?

firecow:
@bcouetil Well, the pipeline and job execution was completely hacked together, and has been like that for a long time.
I've simplified job execution in the JobExecutor class, and added a lot of missing testes, to fix the issues reported.
So its not really a change in philosophy, as much as it is a move towards the current philosophy, with lots and lots of integration testing :smile:

bcouetil:
OK, merge when you want :wink:

PR #165, opened 2021-04-28
Fix: Load variables as string. Fixes #161

### LINKED ISSUE #161: Convert all variables to string
Yaml loader is inferring the type of the values specified in `variables.yaml`  and typecasting the loaded variable value accordingly. This causes the following check to fail when a non-string value is encountered

https://github.com/firecow/gitlab-ci-local/blob/master/src/parser.ts#L136

For example, the following `variables.yaml` file causes `gitlab-ci-local` to fail with error

> TypeError: value.match is not a function
>     at Function.initHomeVariables (/snapshot/gitlab-ci-local/src/parser.ts:136:24)
>     at Parser.init (/snapshot/gitlab-ci-local/src/parser.ts:173:30)
>     at Function.create (/snapshot/gitlab-ci-local/src/parser.ts:48:9)
>     at Object.handler (/snapshot/gitlab-ci-local/src/handler.ts:68:18)
>     at Object.handler (/snapshot/gitlab-ci-local/src/index.ts:35:21)

```
# variables.yaml

project:
  git.myhost.com/packages/packagename.git:
    MY_TEST_VAR: 123
```

JoyceBabu:
Since environment variables are always strings, we should load all variables too as string.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -90,7 +90,7 @@ export class Parser {
             return {};
         }
 
-        const data: any = yaml.load(await fs.readFile(variablesFile, "utf8"));
+        const data: any = yaml.load(await fs.readFile(variablesFile, "utf8"), { schema: yaml.FAILSAFE_SCHEMA });
         let variables: { [key: string]: string } = {};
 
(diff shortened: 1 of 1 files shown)


firecow (approved):
LGTM

PR #167, opened 2021-04-28
Prevent recursive copying of files by rsync. Fixes #166

### LINKED ISSUE #166: Recursive `.docker` directory creation
@firecow The following code is causing an issue when running from the typescript source directly. I did not notice this, when running the binary.

```
rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) . .gitlab-ci-local/builds/.docker/
```

The above command is copying the current directory to a sub directory. But `.gitlab-ci-local` is not ignored in the project's `.gitignore`. So this is causing rsync to copy the files in a recursive loop.

```
joyce@mbp blockhash % mkdir .gitlab-ci-local/builds/.docker
joyce@mbp blockhash % du -sch .gitlab-ci-local/builds/.docker
  0B    .gitlab-ci-local/builds/.docker
  0B    total
joyce@mbp blockhash % rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) . .gitlab-ci-local/builds/.docker/
^Crsync error: received SIGINT, SIGTERM, or SIGHUP (code 20) at io.c(503) [generator=3.2.3]
rsync error: received SIGINT, SIGTERM, or SIGHUP (code 20) at rsync.c(748) [sender=3.2.3]
joyce@mbp blockhash % du -sch .gitlab-ci-local/builds/.docker
534M    .gitlab-ci-local/builds/.docker
534M    total
```
In 5 seconds, the directory has grown to 534MB.

But if I exclude the `.gitlab-ci-local` directory in the `rsync` command, this recursion does not happen.

```
joyce@mbp blockhash % mkdir .gitlab-ci-local/builds/.docker
joyce@mbp blockhash % rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) --exclude .gitlab-ci-local/ . .gitlab-ci-local/builds/.docker/
joyce@mbp blockhash % du -sch .gitlab-ci-local/builds/.docker
104M    .gitlab-ci-local/builds/.docker
104M    total
joyce@mbp blockhash % rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) --exclude .gitlab-ci-local/ . .gitlab-ci-local/builds/.docker/
joyce@mbp blockhash % du -sch .gitlab-ci-local/builds/.docker
104M    .gitlab-ci-local/builds/.docker
104M    total
```

So I think we need to add `--exclude .gitlab-ci-local/` to the above rsync command.`

JoyceBabu:
Since the target path is a sub directory of the source path, rsync is recursing into the copied directory causing an infinite loop.

Fixed by adding the target directory to the `rsync` exclude list.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -134,7 +134,7 @@ export class Utils {
     static async rsyncNonIgnoredFilesToBuilds(cwd: string, target: string): Promise<{ hrdeltatime: [number, number] }> {
         const time = process.hrtime();
         await fs.ensureDir(`${cwd}/.gitlab-ci-local/builds/${target}`);
-        await Utils.spawn(`rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) ./ .gitlab-ci-local/builds/${target}/`, cwd);
+        await Utils.spawn(`rsync -ah --delete --exclude-from=<(git -C . ls-files --exclude-standard -oi --directory) --exclude .gitlab-ci-local/builds/${target}/ . .gitlab-ci-local/builds/${target}/`, cwd);
         return {hrdeltatime: process.hrtime(time)};
     }
 


JoyceBabu:
This only solved part of the issue.

bcouetil:
Could we add the whole tool folder ? `--exclude .gitlab-ci-local`

JoyceBabu:
> Could we add the whole tool folder ? `--exclude .gitlab-ci-local`

I initially did suggest excluding the whole directory.  We need to copy file variables and artifacts from `.gitlab-ci-local` to the docker container. I am not sure how that is handled right now, hence I have limited the exclusion  to the target directory in this PR.

bcouetil:
Those additional copies can be done from the original folder, if thats not already the case. I think it is, since this rsync is very new.

A test would verify this.

bcouetil:
I think this PR is still valuable, and a good performance optimization. I'll let @firecow decide.

firecow:
@bcouetil @JoyceBabu I'll make a little test case, to make sure this actually does something.
I have a hard time seeing it actually does anything at all :smile:

JoyceBabu:
I agree. As mentioned in #166, I am unable to replicate this now. The funny thing is that a few hours ago the same repository was showing this behaviour.

But to confirm that this is an edge case and not to a human error, we need a test case. I will try to create a test repo, if I encounter this again. Until then, I too agree that this can be closed.

PR #168, opened 2021-04-29
Auto generate .gitlab-ci-local/.gitignore file If .gitignore file isn't git add'ed right away, make sure .gitlab-ci-local is excluded from rsync

firecow:
@bcouetil @JoyceBabu 

I was able to reproduce, when ever .gitlab-ci-local/.gitignore wasn't manually git add'ed, the recursive behavior occurs.

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -21,6 +21,14 @@ const checkFolderAndFile = (cwd: string, file?: string) => {
     assert(fs.existsSync(gitlabFilePath), `${cwd} does not contain ${file ?? ".gitlab-ci.yml"}`);
 };
 
+const generateGitIgnore = (cwd: string) => {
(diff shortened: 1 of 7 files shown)


bcouetil:
I came to say that I easily reproduced also :smile:

bcouetil:
Tested this branch, working like a charm :muscle:

firecow:
@bcouetil Approve, lets get it right in :smile:

JoyceBabu:
@firecow Great. Usually when you add a gitignore file with content '*`, you won't be committing the file (unless you want to create an empty directory). So, the PR is useful. But now I think it is safe to exclude the entire `.gitlab-ci-local` directory, not just `.gitlab-ci-local/builds/${target}`.

bcouetil (approved):
tested, LGTM

PR #171, opened 2021-04-29
Make sure to exit with 1 if any jobs ended in FAIL

firecow:


bcouetil (approved):
Nice catch with the allow_failure special case. LGTM

PR #187, opened 2021-05-17
Make sure filename and docker container/volume names use a safe jobnme format extracted via base32 (Crockford)

firecow:


bcouetil (approved):
LGTM

PR #210, opened 2021-07-08
Propagate the --privileged flag to services

lvjp:
Services are working, but the `--privileged` do not affect them.  
This can be bad for docker-in-docker based jobs.

For example, with this `.gitlab-ci.yml` :
```yaml
myJob:
  image: docker:latest
  service:
    - docker:dind
  script:
    - docker build -t myRegistry/myImage:latest
    - docker push myRegistry/myImage:latest
```

The docker service will be launched but rapidly fail with few errors like this one :

> mount: permission denied (are you root?)

With the **gitlab-runner**, when the [privileged mode](https://docs.gitlab.com/runner/executors/docker.html#use-docker-in-docker-with-privileged-mode) is active, the services are also affected.

In a way to fix the dind use-case and to better emulate the **gitlab-runner** behavior, we should propagate the `--privileged` flag to services.

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -403,7 +403,7 @@ export class Job {
                 dockerCmd += `--network gitlab-ci-local-${this.jobId} `;
                 for (const service of this.services) {
                     await this.pullImage(writeStreams, service.getName(this.expandedVariables));
-                    await this.startService(writeStreams, service);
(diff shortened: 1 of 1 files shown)


firecow (approved):
Good job @lvjp

PR #215, opened 2021-07-22
No more enforce manual rule for interactive jobs

lvjp:
Sometimes, you won't run inside a docker because you cannot use the local environment facility, and you use an interactive job to do the job ;-)

But each time you want to run the pipeline, you should add the `--manual myJob` in the command line.  
This is tedious and error-prone because you can easily forget it !  
That why I want to remove the `manual` enforcement in interactive jobs.

My use case :
- **Remote pipeline**
    - **First job**: build a docker image and I send it to my company private docker registry
    - **Second job**: Implicitly fetch image from the private docker registry and run integration tests
- **Local pipeline**
    - **First job**: build a docker image on the local docker daemon
    - **Second job**: Use local docker image and run integration tests

I don't think that there are any drawbacks since you are still able to manually enforce the `manual` rule

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -125,8 +125,8 @@ export class Job {
             this.allowFailure = ruleResult.allowFailure;
         }
 
-        if (this.interactive && (this.when !== "manual" || this.imageName !== null)) {
-            throw new ExitError(`${this.chalkJobName} @Interactive decorator cannot have image: and must be when:manual`);
+        if (this.interactive && this.imageName !== null) {
+            throw new ExitError(`${this.chalkJobName} @Interactive decorator cannot have image:`);
         }
 
         if (this.injectSSHAgent && this.imageName === null) {


firecow:
I'm not sure, I quite understand the use-case....
I use "interactive" jobs, for shooting up an' interactive terminal session.
An interactive terminal session cannot be part of an actual automatic pipeline as far as I can see at least.
Can you describe your "use-case" perhaps by pasting some pseudo .gitlab.yml?
@lvjp

lvjp:
Yes, I do not need an **interactive** session. I just need to directly run on the host machine to take leverage of the local docker setup, that's why I use it.

```yaml
.build:
  script:
    - docker build -t registry.myCompany.local ./src
    - '[ ${GITLAB_CI} == "true" ] && docker push registry.myCompany.local'

build:
  extends: .build
  image: docker
  services:
    - docker:dind

# @Interactive
manual_build:
  extends: .build
  # Since we use an interactive shell, we are not inside a docker and can directly use the host docker setup
  rules:
    - if: $GITLAB_CI == 'false'

test:
  image: docker
  services:
    - docker:dind
  script:
    # On remote, docker will implicitly do a docker pull prior to running the image
    # On local, docker will use the previously builded image since it's still in the docker local cache
    - docker run --rm registry.myCompany.local /run/myTests.sh
```

firecow:
@lvjp Ah, I understand now...
The `@Interactive` decorator is not ment to force the session away from docker into shell.
Its purpose is to keep keep stdin open for user inputs, no matter if the job has `image:` or not.
We just haven't implemented jobs with `image:` to work with `@Interactive` yet.

Can't you just solve your "local-only" needs like this?
```
---
.build:
  script: docker build -t imagename src/

.push:
  script: docker push imagename

webserver-image-local:
  stage: build
  rules:
    - if: $GITLAB_CI == 'false'
  script:
    - !reference [.build, script]

webserver-image:
  stage: build
  image: docker:stable
  services: [docker:dind]
  rules:
    - if: $GITLAB_CI == 'true'
  script:
    - !reference [.build, script]
    - !reference [.push, script]
```

lvjp:
Thank you for your answer, it's work well with your corrected example.

I was miss understanding the usage/purpose of the `@Interactive` decorator.  
The `shell` executor usage is only activated because the `image:` is not defined...

firecow:
@lvjp Yes, the lack of `image:` will force the script execution onto the host system.

PR #221, opened 2021-07-27
Custom home variables can now overwrite GitLab predefined ones

### LINKED ISSUE #220: User variable values should take precedence over predefined variable values
As a user I should be able to overwrite the default predefined variables by re-defining them in my `~/.gitlab-ci-local/variables.yml` file.

e.g. 

The predefined variable `$CI_PROECT_ID` has a default value of `1217`. I should be able to overwrite that value with my own by adding the following to my `~/.gitlab-ci-local/variables.yml` file.

```yaml
project:
  gitlab.com/my-namespace/my-project.git:
    CI_PROJECT_ID: 12345678
```

If you agree with this behavior, could it be fixed by simply modifying the order of inputed arrays at [job.ts, line 122](https://github.com/firecow/gitlab-ci-local/blob/8a9d2b4e67737623a74b8a16752260fa241a961c/src/job.ts#L122)?

vinman1:
As a user I should be able to overwrite the default predefined variables by re-defining them in my ~/.gitlab-ci-local/variables.yml file.

Closes #220.

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -119,7 +119,7 @@ export class Job {
         const expandedGlobalVariables = Utils.expandVariables(globals.variables || {}, envs);
         const expandedJobVariables = Utils.expandVariables(jobData.variables || {}, envs);
 
-        this.expandedVariables = {...expandedGlobalVariables, ...expandedJobVariables, ...homeVariables, ...predefinedVariables};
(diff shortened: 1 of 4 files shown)


firecow (approved):
Looks awesome, going into master

PR #239, opened 2021-09-22
Fix PROJECT_PATH generation

lvjp:
`PROJECT_PATH` variable is camel case formatted, even if GitLab do not apply formatting.

I found this bug with a project named `camel-case` which translate in `PROJECT_PATH=camelCase`,
but in a real GitLab instance it was `PROJECT_PATH=camel-case`


firecow (approved):
LGTM

PR #261, opened 2021-10-05
Less strict darwin check

firecow:


src/job.ts
```diff
@@ -366,7 +366,7 @@ export class Job {
         if (!this.injectSSHAgent) {
             return "";
         }
-        if (process.env.OSTYPE === "darwin") {
+        if (process.env.OSTYPE?.match(/^darwin/) ?? null) {
```
atc-cegodk, line 369: Use `process.platform` instead https://nodejs.org/api/process.html#process_process_platform

atli-c: FYI: `process.env.OSTYPE` was not available (on my machine) unless I ran `export OSTYPE` before running `gitlab-ci-local`.

firecow: Done @atc-cegodk

Issue #244, opened 2021-09-27
Question: fetch raw files from private Gitlab instances is possible ?


Neonox31:
First, thanks for this amazing tool 👍

I'm trying to get gitlab-ci-local working with a private Gitlab instance from my company and it results with `Project include could not be fetched` error.
Looking at the code reveals that no private token are handled as an header to fetch raw files and I presume this is the problem, am I right ? Maybe I missed a thing ?

Thanks.

firecow:
Hmmm... How does your include field look? @Neonox31

Neonox31:
@firecow here is the include section :
```
include:
  project: "factory/ci-centre/ci-helpers"
  file: "/templates/.gitlab-ci-tpl.yml"
  ref: "2.10.0"
```
`factory` is a group at top level of our Gitlab self-hosted instance and because project are not publicly visible we need to use a personal token to access them.
I'm thinking of an environment variable that could contain the token and that would be injected in the node-fetch headers when it is set, what do you think ?

firecow:
Ahaa.... That should be doable, what would be a good name constant for such an access token...

Neonox31:
What do you thin of `GCIL_TOKEN` alias Gitlab C.I Local Token ?

firecow:
Hmmm.... I just realized @Neonox31 
```
include:
  project: "factory/ci-centre/ci-helpers"
  file: "/templates/.gitlab-ci-tpl.yml"
  ref: "2.10.0"
```
Are fetched via `git archive` which use ssh private key as authentication method.
Are you sure, that you have ssh access to that project ?

Neonox31:
Oh yes you're right @firecow, sorry !

I understood where my problem came from by trying the command executed by the spawn on my machine and the problem comes from the `tar` command.
On my machine I have _bsdtar_ which requires the file path to be specified (in this case stdin) otherwise it defaults to `/dev/st0`:
![image](https://user-images.githubusercontent.com/1135958/135516857-bdb47475-164c-4b23-ae38-3e0983b79ffd.png)

The workaround is to add `-f -` parameter to `tar command`, the problem is well described here: https://unix.stackexchange.com/a/131053

Would you like a P.R with the fix ?

firecow:
@Neonox31 Ofcourse PR away :smile:

PR #263, opened 2021-10-06
fix(include project file): define file input parameter for better-compatibility on tar command

Neonox31:
https://github.com/firecow/gitlab-ci-local/issues/244#issuecomment-931595348


firecow (approved):
LGTM

PR #290, opened 2021-10-21
No artifacts to source

### LINKED ISSUE #249: artifacts are stored in wrong places since 4.23.0
hi Mads,

i haven't had much time lately to play around with gitlab-ci-local. But today i updated to the newest version from a version somewhere in 4.19.x. I recognized some problems with the artifacts. Subsequent jobs can't access the artifacts, as it was possible before.

I have seen some change in the handling in .gitlab-ci-loca/ folder, now artifacts are stored with safeJobName as a subfolder instead of being stored directly in .gitlab-ci-loca/artifacts. So every job has its own subfolder lije
.gitlab-ci-loca/artifacts/build/path/to/deep/folder
.gitlab-ci-loca/artifacts/test/path/to/deep/folder
...

Another thing that might not be work as desired: The artifacts are also stored in the code repository. With your artifacts-docker testcase i get this result:
myCodeFolder/.gitlab-ci-local
myCodeFolder/.gitlab-ci-local/artifacts/build/path/to/deep/folder
myCodeFolder/path/to/deep/folder

The "path/to/deep/folder" folder is now inside of my code repository folder.

I now use version 4.22.0 as a workaround, so everything seems to work fine with that version

firecow:
CLOSES #249


alexander-matthiesen:
Would it be possible to create a flag to make it global and without touching the pipeline code with decorators?

Something like `gitlab-ci-local --no-artifacts-to-source`

PR #294, opened 2021-10-24
Harden regex for git branch merging

aepfli:
Hardened the logic for merging git branches.

It now looks more descriptive for hashes (no whitespace characters),
it allows to work with generated merges for CI Pipelines,
and it unifies both regexes into one.

I am not sure how to write tests for that, therefore i created a draft Pullrequest, also to check with CI Builds


aepfli:
i added now tests, to verify my git log output - i am not sure if that is the best way to write tests with nodejs :D - i hope this is the right format.

Also they might need some additional tweaking, i am currently just counted the `spawn` calls and on the third i return the log

aepfli:
i was just thinking is `git` a must have - theoretically we could also add fallback values instead of failing the code. this way gitlab-ci-local can be even used in non-git directories/environments. i might miss some invocation regarding that. but it was a thought which popped into my head yesterday ;)

firecow:
@aepfli I don't see anything wrong, with "faking" values extracted via `git` commands. The less mandatory dependencies, the better :smile: 
Perhaps add a warning in the beginning 
```
> Warning: This is not a git repo, certain values will be faked.
```

aepfli:
@firecow i added now some tests, a fallback and a lot of test cases to gitData

I still have some failing tests, but those are also happening in the main branch - it seems like a local setup issue. If you think this changes are good, and they do look sufficient, i will mark this PR as ready. if not i am open for more feedback

firecow:
@aepfli Mark it ready... 
Me and @nbj found another issue with :port in CI_SERVER_URL... Maybe this will take care of that...

firecow (approved):
looks very good

firecow:
I'll merge this in master, to make sure you get the contribution metrics.
I will change a few things once in master, to make it pass the Github workflows, hope that ok...

aepfli:
:+1: fine from myside thank you for your efforts!

PR #309, opened 2021-11-11
Makes the port from gitconfig url part of the domain regex group

nbj:


tests/test-cases/predefined-variables/.gitconfig
```diff
@@ -1,3 +1,3 @@
 
 [remote "origin"]
-	url = git@gitlab.com/gcl/predefined-variables.git
```
aepfli, line 3: i am not sure we still need this file - with my latest pull request which got merged by @firecow - i changed it to use `git remote -v` instead of the file. hence that this file is overall deprecated. or at least those lines :) but please tell me if i am wrong regrading my assumption

src/git-data.ts
```diff
@@ -82,7 +82,7 @@ export class GitData {
         } else {
             throw new ExitError("Could not locate.gitconfig or .git/config file");
         }
-        const gitRemoteMatch = gitConfig.match(/url = .*(?:http[s]?:\/\/|@)(?<domain>.*?)[:|/](?<group>.*)\/(?<project>.*?)(?:\r?\n|\.git)/);
+        const gitRemoteMatch = gitConfig.match(/url = .*(?:http[s]?:\/\/|@)(?<domain>.*)[:|/](?<group>.*)\/(?<project>.*?)(?:\r?\n|\.git)/);
```
aepfli, line 85: i am not sure this regex changes fixes all cases - i would go with something like 
```
/.*(?:http[s]?:\/\/|@)(?<domain>[^:/]*)((:\d+\/)|:|\/){1}(?<group>.*)\/(?<project>.*?)(?:\r?\n|\.git)/
```

i adapted also the current test to be better understandable, so if you rebase onto the latest main and adapt the the `git remote only test` to something like the following - this verifies against a lot of combinations and can be easily adapted

```javascript
   describe("only for git remote data", () => {
        describe("for working input", () => {
            test.each([
                `git@${remote.domain}:${remote.group}/${remote.project}.git`,
                `git@${remote.domain}/${remote.group}/${remote.project}.git`,
                `git@${remote.domain}:222/${remote.group}/${remote.project}.git`,
                `https://git:git@${remote.domain}/${remote.group}/${remote.project}.git`,
                `https://git:git@${remote.domain}:22/${remote.group}/${remote.project}.git`,
            ])("%s", async (url) => {
                when(mock)
                    .calledWith("git remote -v", expect.any(String))
                    .mockReturnValue({
                        stdout: `
                    origin	${url} (fetch)
                    origin	${url} (push)
                    `
                    })
                Utils.spawn = mock
                const gitData = await GitData.getRemoteData("./");
                expect(gitData).toEqual(remote)
            })
        })
```

tests/test-cases/predefined-variables/.gitconfig
firecow: I'm about to remove all .gitconfig's from all test folders.
I just need to figure out a way to make my jest-when mock fallback to the default behavior of Utils.spawn
You wouldn't happen to know, how this is done, would ya @aepfli ?

```typescript
// This is pseudo
import {when} from "jest-when";

export function initSpawnMock(spawnMocks: {cmd: string; returnValue: any}[]) {
    const mock = jest.fn();

    for (const spawnMock of spawnMocks) {
        when(mock).calledWith(spawnMock.cmd, expect.any(String)).mockResolvedValue(spawnMock.returnValue);
    }

    when(mock).calledWith(expect.any(String), expect.any(String)).mockResolvedValue( !!! default old Utils.spawn here !!!!)

    return mock;
}
```

aepfli: i would go with a spy here - based on my java knowledge - i need to check how jest-when is working with a spy, but you normally want to have the original implementation, and just overwrite certain calls.

aepfli:
```typescript
// This is pseudo
import {when} from "jest-when";
import {Utils} from "...";

export function initSpawnMock(spawnMocks: {cmd: string; returnValue: any}[]) {
    const mock = jest.spyOn(Utils, "spawn");

    for (const spawnMock of spawnMocks) {
        when(mock).calledWith(spawnMock.cmd, expect.any(String)).mockResolvedValue(spawnMock.returnValue);
    }

    return mock;
}
```
something like that - not tested pseudo code

PR #312, opened 2021-11-15
GitHub Actions for Pull Requests

aepfli:
Currently it is not possible to run the tests via GitHub Actions,
if it is not happening on the main repository. This change should
allow contributors to leverage GitHub Actions within their Pull
Requests. This should increase the quality of contributions, and
allow better focus for the maintainers.

it can be seen in action here: https://github.com/aepfli/gitlab-ci-local/pull/2
- sonarcloud will fail here, as i do not have it within my repo, but in the future, it should be disabled when I am forming ;)
### Notes: 

- this works because i pointed some tests to my fork on gitlab - there is some manual action needed (adding a deploy key with read permissions, so this can work)

- I reduced it to be not running for all branches, but for all PRs
- i removed the depth for sonarcloud, as i do not think it is needed to know, who has created an issue .
- codeQL is reporting an issue, because we are now not building for all branches
- the ssh key and the known hosts are now stored in files within the project, as they only have read permissions and are only used for testing, this should be fine
- yamllint running locally is good, but there is also an action, and if each change is introduced via pull request, there is more safety and it should be checked anyways

### Sidenote

I felt really bad to introduce so many linting errors, and therefore i thought this might be a good addition. there are other thoughts also in my head, which could make the future development easier. but one after the other. i am open for feedback.


firecow:
This is pure gold, lets get it in @aepfli

firecow (approved):
LGTM

firecow:
@aepfli I'll merge it, if you say go

tests/test-cases/include-remote-with-inner-local/integration.include-remote-with-inner-local.test.ts
```diff
@@ -8,7 +8,7 @@ test("include-remote-with-inner-local", async () => {
     const writeStreams = new MockWriteStreams();
     const spyGitRemote = {
         cmd: "git remote -v",
-        returnValue: {stdout: "origin\tgit@gitlab.com:gcl/include-remote-with-inner-local.git (fetch)\norigin\tgit@gitlab.com:gcl/include-remote-with-inner-local.git (push)\n"},
+        returnValue: {stdout: "origin\tgit@gitlab.com:aepfli/include-remote-with-inner-local.git (fetch)\norigin\tgit@gitlab.com:aepfli/include-remote-with-inner-local.git (push)\n"},
```
aepfli, line 11: @firecow should this be gcl? or should this be firecow? i do not have access to gitlab.com/gcl :) if we leave it with `gcl` this will most likely affect everyone. so i suggest to either set it to `firecow` or make `gcl` public accessible

aepfli:
before we merge, we should get the tests back to the original status, and ensure that the deploy key is setup properly within your gitlab.com/firecow/gitlab-ci-local-includes or `gcl` depending on what we should use :)

you can find that setting here: 
https://gitlab.com/firecow/gitlab-ci-local-includes/-/settings/repository#js-deploy-keys-settings

image

tests/test-cases/include-remote-with-inner-local/integration.include-remote-with-inner-local.test.ts
firecow: I'm just mocking the remote host with the gcl thingie...
It's still fetching from gitlab.com/firecow/gitlab-ci-local-includes
There is no need to change this test or its matching .gitlab-ci.yml

aepfli: well if i read the code correctly this will end up in a remote like `gitlab.com:gcl/firecow/gitlab-ci-local-includes` which is an url like `gitlab.com/gcl/firecow.gitlab-ci-local-includes` (gitlab also allows group names as ports) 

seen here
https://github.com/firecow/gitlab-ci-local/blob/master/src/parser-includes.ts#L120

tests/test-cases/include-remote-with-inner-local/.gitlab-ci.yml
```diff
@@ -1,3 +1,3 @@
 ---
 include:
-  - { project: firecow/gitlab-ci-local-includes, file: .gitlab-module-with-local.yml, ref: include-string-list }
+  - { project: aepfli/gitlab-ci-local-includes, file: .gitlab-module-with-local.yml, ref: include-string-list }
```
aepfli, line 3:
```suggestion
  - { project: firecow/gitlab-ci-local-includes, file: .gitlab-module-with-local.yml, ref: include-string-list }
```

tests/test-cases/include-remote-with-inner-local/integration.include-remote-with-inner-local.test.ts
aepfli:
```suggestion
        returnValue: {stdout: "origin\tgit@gitlab.com:firecow/include-remote-with-inner-local.git (fetch)\norigin\tgit@gitlab.com:firecow/include-remote-with-inner-local.git (push)\n"},
```

aepfli: but i realize this should than also work with other tests... maybe i derped

aepfli:
i changed the paths back - if the deploy token is in place, the test should run through, and we should be ready to merge :)

aepfli:
Looks, good we can get this merge :) - if there are issues, let me know, i am glad to support :) but i can not test everything within my fork :)

firecow:
Lets go :D

PR #316, opened 2021-11-17
Github Actions. Sonar cloud run from main repo via pull_request_target

firecow:


aepfli:
yes i see the error with the if clause and the branch - something like
```
if: ${{ github.event.pull_request.head.repo.full_name == github.repository }} ||  {{github.event_name != 'pull_request'}}
```

i can investigate and provide a proper solution

firecow:
@aepfli https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/ Do you read this the same way that I do ?

firecow:
I wonder how dangerous SONAR_CLOUD_TOKEN would be, if placed in the source code. I shall read.

aepfli:
an alternativ to bypass this all would be a second github account. this account could be added to this project on sonarcloud, and only granted analysis permissions:

image

this way the secret for sonarcloud is not as critical, and it could be used by everyone

aepfli:
i have another project where we placed it in the source code - it is actually mine, and that means that people can mess with my projects, can administer issues etc. It is not like they can change some code etc. but they can mess with the sonarcloud

firecow:
I think pull_request_target would be fine... pull_requests from forks, won't be able to run modified `build.yml` then..

Issue #301, opened 2021-10-29
docker:dind service failure


niki9796dk:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
stages:
  - build

build-dev:
  image: docker:latest
  stage: build
  services:
    - docker:dind
  script:
    - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD registry.gitlab.com
```

**Expected behavior**
https://gitlab.com/niki9796dk/gitlab-ci-local-debugging/-/jobs/1731462426

**Host information**
Ubuntu 20.04.2 LTS
gitlab-ci-local 4.25.1

**Additional context**
gitlab ci local console dump
```
parsing and downloads finished in 55 ms
build               starting docker:latest (build)
build               pulled docker:latest in 61 ms
build               pulled docker:dind in 66 ms
build               started service image: docker:dind with aliases: docker, docker in 567 ms
build               copied to container in 1.36 s
build               copied artifacts to container in 1.91 s
build               $ docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD registry.gitlab.com
build               > WARNING! Using --password via the CLI is insecure. Use --password-stdin.
build               > still running...
build               > error during connect: Post "http://docker:2375/v1.24/auth": dial tcp: lookup docker: Try again
build               finished in 26 s  FAIL 1  

 FAIL  build  
```

firecow:
This is most likely todo with missing healthchecking

ryall:
I'm seeing the same, I think:

```
build_image $ docker build . -t ${CI_REGISTRY_IMAGE}:${CI_COMMIT_TAG}
build_image > time="2021-11-07T15:30:19Z" level=error msg="failed to dial gRPC: cannot connect to the Docker daemon. Is 'docker daemon' running on this host?: dial tcp: lookup docker on 127.0.0.11:53: no such host"
[...]

PR #326, opened 2021-11-29
Add serviceHealthCheck

HeWhoWas:
This PR is an amateur attempt at adding a generic service healthCheck function. It works by:

1. Using `docker inspect <service_name>` to identify a list of ports exposed by the image.
2. Runs an instance of [willwill/wait-for-it](https://hub.docker.com/r/willwill/wait-for-it) in docker, that wait for up to 30 seconds for the service to become available.

This logic was based on a small blurb on the [GitLab Services Page](https://docs.gitlab.com/ee/ci/services/#how-the-health-check-of-services-works)

> To make sure this works, the runner:
>
>    1. Checks which ports are exposed from the container by default.
>    2. Starts a special container that waits for these ports to be accessible. 

I'm not a JS developer, and this is literally the first time I've used TS. If I've done anything egregious, happy to change it.

Should fix https://github.com/firecow/gitlab-ci-local/issues/301

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -460,6 +460,7 @@ export class Job {
                 for (const service of this.services) {
                     await this.pullImage(writeStreams, service.getName(this.expandedVariables));
                     await this.startService(writeStreams, service, privileged);
+                    await this.serviceHealthCheck(writeStreams, service);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


firecow:
@HeWhoWas All in all good work... 

Lets use https://www.npmjs.com/package/wait-port instead of "eating" the overhead of starting a container just to check if a TCP port is responding.

HeWhoWas:
I originally intended to use the `net` package in order to introduce as few dependencies as possible, but it didn't work due to the fact that the service name only resolves within the same docker network. If we want to avoid calling out to `docker run` then I think the only option would be to expose the docker DNS server to the host and query it directly.

firecow:
@HeWhoWas Thats right, didn't think about docker network dns name... My bad :smile:

firecow (approved):
lgtm

HeWhoWas:
No problem, thanks!

PR #344, opened 2021-12-16
Print last three lines on script failure or warning

### LINKED ISSUE #342: Print the last [x] lines for fail/warn jobs in end report for easier failure debugging
image

Here, we have no chance of seeing the actual problem, unless we start a painful up scrolling workflow.
It would be nice that the last x lines for `docker-image` was printed in the end report

firecow:
CLOSES #342


moberghammer (approved):
LGTM

PR #347, opened 2021-12-18
--fetch-includes option

firecow:
This MR makes all include be fetched once, and only once.

This increases speed on reruns of the pipeline locally, but will require a manual action to refetch includes from external sources using `--fetch-includes` 

Also simplified `GitData` quite a bit.


moberghammer (approved):
LGTM

PR #367, opened 2022-01-17
Draft: Fix CommitData REF_NAME

FnuGk:
current branch can be set to the wrong value if multiple branches is at the same sha


firecow:
@FnuGk This should be fixed in 4.29.0

PR #422, opened 2022-04-06
Update README.md

FaisalAfroz:
Fix typo for .env filetype

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ curl -L https://github.com/firecow/gitlab-ci-local/releases/latest/download/win.
 
 ### DotEnv file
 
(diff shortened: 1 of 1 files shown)


firecow:
Its not a typo 😁
The code actually looks for a file with that name.

FaisalAfroz:
I got that but a dotenv file usually ends with .env so it is easier for us to use intellisense and we usually .gitignore these env variables so that secrets does not get commited to the source repo.

firecow:
Right, add it to https://github.com/firecow/gitlab-ci-local/discussions/335 its a breaking change, suited to 5.x.x

PR #459, opened 2022-05-06
apt should only work on Debian-based distros

alexislefebvre:


firecow:
What distro are you trying out?

firecow (approved):
lgtm

alexislefebvre:
I didn't try to use it but I'm pretty sure that distros based on Red Hat or Arch Linux (to name a few) won't be provide `apt`.

PR #461, opened 2022-05-06
fix #452: rsync exludes files that it should not

### LINKED ISSUE #452: rsync exludes files that it should not
In the [rsync command][0] to send files to the container there is unexpected behavior.

``` bash
rsync -a --delete-excluded --delete --exclude-from=<(git ls-files -o --directory) --exclude .gitlab-ci-local/ ./ .gitlab-ci-local/builds/${target}/
```

Relevant excerpts from `rsync` docs:

> if the pattern starts with a / then it is anchored to a particular spot in the hierarchy of files, otherwise it is matched against the end of the pathname.

So if `git ls-files -o --directory` outputs "test" because there's a folder at the root of the project that is ignored, rsync will also ignore any path ending in "test" as well, leading to unintended and confusing behavior.

One additional example is with golang, where often `./cmd/mybinaryname` is a directory in the project, and the `.gitignore` contains `mybinaryname` to avoid committing the binary. The result is that rsync will fail to sync the folder `./cmd/mybinaryname`.

Unfortunately I haven't had time to try and patch it, but it looks like prefixing the output of `git ls-files -o --directory` with a `/` may be sufficient.

[0]: https://github.com/firecow/gitlab-ci-local/blob/5058be54dcb41d0b66938603a8a0f3f97142e711/src/utils.ts#L143

icook:
see #452

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -140,7 +140,7 @@ export class Utils {
     static async rsyncTrackedFiles(cwd: string, target: string): Promise<{ hrdeltatime: [number, number] }> {
         const time = process.hrtime();
         await fs.mkdirp(`${cwd}/.gitlab-ci-local/builds/${target}`);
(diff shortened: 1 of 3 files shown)


firecow (approved):
lgtm

PR #480, opened 2022-05-22
Make it possible to specify relative path to .gitlab-ci-local state directory

### LINKED ISSUE #16: Configure gitlab-ci-local per project via .gitlab-ci.local/config.yml

### LINKED ISSUE #475: Configurable `.gitlab-ci-local` directory
When running the pipeline the directory `.gitlab-ci-local` is created in the root of the project

This appears to be hardcoded as `{cwd}/.gitlab-ci-local` (where cwd is configurable via --cwd)

Allowing a configuration of this path would make adoption easier:
- Symfony has a default tmp dir: `{Symfony.root}/var/tmp/`
- RubyOnRails has a default tmp dir: `"#{Rails.root}/tmp/"`

In our project we have a general tmp directory for various ci tools and we want to add `.gitlab-ci-local` to that path.

firecow:
closes #475


LauJosefsen (changes requested):
If the configurable path should be able to be absolute, I would suggest resolving the path with cwd initially, and not do `cwd/stateDir` as that would make it unable to be absolute paths.

firecow:
This feature is relative to CWD only, stated in index.ts

LauJosefsen (approved):

Issue #298, opened 2021-10-28
Error: "/gcl-scripts/" could not be found on container with docker-type runner


ben-alkov:
docker-type runner errors out with
```sh
gitlab-ci-local
parsing and downloads finished in 57 ms
uname starting registry.fedoraproject.org/fedora:34 (do_stuff)
uname pulled registry.fedoraproject.org/fedora:34 in 231 ms
uname copied to container in 365 ms
Error: "/gcl-scripts/" could not be found on container 4615c86204aa0eaecb04a4746a644ff1d6eca4c4255b18650068370949f87079: No suchfile or directory
```

**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
uname:
  image: registry.fedoraproject.org/fedora:34
  script:
    - uname -a
```

**Expected behavior**
No *public* url available.

**Host information**
Fedora 34
gitlab-ci-local 4.26.0

**Additional context**
I can't get e.g. `gitlab-runner exec docker uname` to work, as 'gitlab-runner' *requires* (seems to?) a "served" git repo to pull from (i.e. it ignores CWD, saying "fatal: 'FOO' does not appear to be a git repository. fatal: Could not read from remote repository").

Providing the actual upstream repo URL (using `--clone-url`) doesn't work due to cert issues (and neither `--docker-tlsverify=no` nor `variables: GIT_SSL_NO_VERIFY: "true"` in '.gitlabc-ci.yml' make any difference), which I'm pretty sure I have no hope of figuring out... which is why I'm here :)

firecow:
```yml
---
uname:
  image: debian:stable-slim
  script:
    - uname -a
```
@ben-alkov 
Put this in a `.gitlab.yml` and run `gitlab-ci-local`
Paste the output here.

firecow:
@ben-alkov I'm closing this for now..

Feel free to reopen, when you got time to run my `.gitlab-ci.yml` above

ben-alkov:
Part of this is a combination of correctly configuring podman's docker socket "emulation", part of it is SELinux perms on the repo directories under test. With both of those things configured correctly, I am able to get `gitlab-runner exec docker uname` to work on the local repo as expected (frickin' sweet!).

I haven't had a chance to try with a different image, and given the above, I haven't re-tested with gitlab-ci-local

firecow:
I have no idea why `gitlab-runner exec docker uname` is being involved :smile: 
In fact I have no idea what your are trying to do.
[...]

PR #492, opened 2022-05-26
Use generated gcl-entry script to better comply with entrypoint syntax

### LINKED ISSUE #206: Override "entrypoint" cause on odd error
I'll be honest here, I've done a little investigating, but this one is beyond me.

Using a build step that defines a service with an entrypoint, I received this error:
```
services:entrypoint must be an array
```

But, my entrypoint is most definitely an array:
```yaml
      entrypoint:
        - 'docker-entrypoint.sh'
        - '-c'
        - 'fsync=off'
        - '-c'
        - 'shared_buffers=256MB'
```

(I originally used the compact array syntax here to no avail)

I added some logging to src/service.ts:
```typescript
@ src/service.ts:21 @ export class Service {
        if (!this.serviceData || !this.serviceData.entrypoint) {
            return null;
        }
        console.log(this.serviceData);
        console.log(this.serviceData.entrypoint);
        console.log(typeof this.serviceData.entrypoint);
        assert(typeof this.serviceData.entrypoint !== "object", "services:entrypoint must be an array");
        return this.serviceData.entrypoint;
    }
```

and the resulting output:
```typescript
{
  name: 'xxx:latest',
  entrypoint: [
    'docker-entrypoint.sh',
    '-c',
    'fsync=off',
    '-c',
    'shared_buffers=256MB'
  ],
  command: undefined,
  alias: undefined
}
[
  'docker-entrypoint.sh',
  '-c',
  'fsync=off',
  '-c',
  'shared_buffers=256MB'
]
object
```

So I commented out the assertion and was greeted with:
```
docker: Error response from daemon: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "shared_buffers=256MB": executable file not found in $PATH: unknown.
```

firecow:
closes #206 

remedy issue with podman https://github.com/firecow/gitlab-ci-local/issues/298

refactor pull image function to avoid `Utils.bash` and `docker image ls` slowness

### DIFF
--- a/src/job-expanders.ts
+++ b/src/job-expanders.ts
@@ -77,8 +77,7 @@ export function services(gitlabData: any) {
                     entrypoint: expandedService.entrypoint,
                     command: expandedService.command,
                     alias: expandedService.alias,
-
-                });
+                }, Number(index));
             }
         }
     });
--- a/src/job.ts
+++ b/src/job.ts
@@ -926,6 +926,7 @@ export class Job {
 
     private async startService(writeStreams: WriteStreams, service: Service) {
         const cwd = this.argv.cwd;
+        const stateDir = this.argv.stateDir;
         let dockerCmd = `docker create -u 0:0 -i --network gitlab-ci-local-${this.jobId} `;
         this.refreshLongRunningSilentTimeout(writeStreams);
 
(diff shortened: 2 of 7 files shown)


moberghammer (approved):
LGTM

PR #493, opened 2022-05-27
Service healthchecking was broken, apparently!!

### LINKED ISSUE #432: Multiple aliased service declarations will never pass healthchecks.
Example yaml chunk:

```yaml
services:
    - name: postgres:14
      alias: postgresql
    - name: postgres:14
      alias: postgresql2
```

See: https://github.com/firecow/gitlab-ci-local/blob/ff514e5b8acb56d4c677e9b6e0630dd13f158a70/src/job.ts#L981

Because `aliases[0]` is the only hostname checked, the healthcheck will always fail. `./wait-for-it.sh` will look for `postgres:5432`, which doesn't exist. You can work around this by re-tagging the docker image locally (e.g. `docker tag postgres:14 postgresql:14`), which allows you to instead use the image name `postgresql`, giving it the correct hostname to check for.

firecow:
closes #432

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -975,9 +975,7 @@ export class Job {
     }
 
     private async serviceHealthCheck(writeStreams: WriteStreams, service: Service, containerId: string) {
-        const cwd = this.argv.cwd;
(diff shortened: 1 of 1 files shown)


LauJosefsen (approved):
LGTM

PR #536, opened 2022-07-05
Artifacts no-shell-isolation / image mix

firecow:
Jobs generating artfacts in no-shell-isolation now copies their artifacts to stateDir/artifacts folder, so image and shell isolated jobs can use them.


alexander-matthiesen:
I do not really understand what the feature "--no-shell-isolation" is used for. Is this a GitLab feature?

firecow:
Shell isolation = true is the default for jobs without image:

alexander-matthiesen:
Ah okay, I get it. So the shell executor is basically a job that is acting like it is a command on the gitlab runner right? That makes sense to "store" the created files somewhere then.
Do you have a usecase where you use this feature? I have never come to the point where I need a file created on the gitlab runner itself. In three years we always worked with artifacts.

It's looking good from my point of view.

alexander-matthiesen (approved):
lgtm

firecow:
Yeah, we use it quite extensively...
Even though docker-executor should be the best choise.
But we just started off with shell and stuck too it.

Issue #554, opened 2022-08-06
supporting "exists" feature in rules


alexander-matthiesen:
**Is your feature request related to a problem? Please describe.**
Currently the [exists feature from gitlab ci](https://docs.gitlab.com/ee/ci/yaml/#rulesexists) is not supported. That means even if the job is correctly configured and e.g.  should only run if a folder "/docs" exists in the root of the repository but it doesn't, it will be executed regardless.

**Describe the solution you'd like**
The functionality has to be added in the job.ts where we have to handle it in the rules section. Therefore we need to read out the exists configuration after checking the "if: string". If there is an exists configuration we have to check if the folder / file is present. If that is not the case, the job needs to be skipped, if that is the case,  the job should be executed.

**Describe alternatives you've considered**
I have no alternative to the solution above.

**Additional context**
This could be the base of the "changes" feature where we additionally need to check if the current git status is the same as when the file was changed.

PR #555, opened 2022-08-07
feat(554): supporting exists feature in rules

alexander-matthiesen:
#554


alexander-matthiesen:
I've noticed that the github action "build/jest" is always checking out master. Therefore the added tests for rules exists are not taken please in the "pipeline". It might be confusing since every action is green and might suggest a working test testsetup.

firecow:
@alexander-matthiesen
Can you spot my mistake in .github/workflows?

alexander-matthiesen:
I've tried a bit but unfortunately I haven't that much experience with github workflows. I'm more of an expert with GitLab CI/CD (thats also why I'm here. :D 

I will have a look. Maybe it is working better now when I can create a branch in this repo here instead of having a fork.

firecow:
Hehe, me too 😁I'll read up, and try some things, then you can pull master and see if that helps 🤟

alexander-matthiesen:
Actually I will close this PR and create a new one based on a branch in this repository. 
I really don't like the "forking" stuff.
Follow up PR will come asap.

PR #556, opened 2022-08-07
feat(554): supporting rules exists

### LINKED ISSUE #554: supporting "exists" feature in rules

alexander-matthiesen:
Closes #554


alexander-matthiesen:
@firecow I think I got it.

The issue was, that we had either "push.branched: master" or "pull_request_target" (Basically master again) but we need just "pull_request" like described here: https://stackoverflow.com/a/58197837

alexander-matthiesen:
@firecow , this should work now. :)

firecow (approved):
lgtm

PR #558, opened 2022-08-08
feat(197): Adding tests for include wildcard

### LINKED ISSUE #197: wildcards when including local files
https://about.gitlab.com/releases/2021/05/22/gitlab-13-12-released/#support-wildcards-when-including-yaml-cicd-configuration-files

alexander-matthiesen:
Closes #197

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@
 /coverage
 
 !/docker-npm
+
+/.gitlab-ci-local/
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "execa": "5.1.1",
         "fs-extra": "10.1.0",
         "js-yaml": "4.1.0",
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -13,3 +13,5 @@
 /coverage
 
 !/docker-npm
+
+/.gitlab-ci-local/
```
firecow, line 17: Wouldn't we like to know, if the tool by accident creates a .gitlab-ci-local folder ?

alexander-matthiesen: I would not check this via git because it might be that someone will commit a .gitlab-ci-local folder which is not what we want I guess. 

So we could create a test to check if a .gitlab-ci-local folder was created or add it in the package.json file in the end of the test command.

firecow: Yah, you are right...

firecow (approved):
lgtm

PR #564, opened 2022-08-12
Jobs that gets when:never because of rules.exist is shown in --list

firecow:
- Make sure rules.exist is evaluated on parse time, so `--list` and `--list-all` can show proper when value
- Use micromatch over minimatch

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,8 @@
       "@typescript-eslint/indent": [
         "error",
         4
-      ]
+      ],
+      "space-in-parens": "error"
     }
   },
   "jest": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -31,7 +31,7 @@
     "execa": "5.1.1",
     "fs-extra": "10.1.0",
     "js-yaml": "4.1.0",
-    "minimatch": "5.1.0",
+    "micromatch": "^4.0.5",
```
alexander-matthiesen, line 34: What's the difference between those two? In my research I just found minimatch. Didn't know there was an alternative.

alexander-matthiesen (approved):

package.json
firecow: More features and it performs better than minimatch

firecow:
@alexander-matthiesen I've hopefully just fixed the issue :smile:

PR #619, opened 2022-10-14
no-shell artifacts bugfix

firecow:
- Fix a bug, where files inside `.gitlab-ci-local` could be copied out as an artifact
- Simplify rsync command in `copyArtifactsOut` function


atli-c (approved):
LGTM

PR #621, opened 2022-10-14
Always put service container logs in .gitlab-ci-local/services-output

### LINKED ISSUE #611: Add output for services
A recurring issue using Gitlab is the lack of visibility about services. It is quite impossible a get logs from services for debug.  
There is the same issue using gitlab-ci-local. Services are removed at the end of job and no logs are kept.

I would be great if services logs could be copied/kept in `.gitlab-ci-local/output` for futher analysis.

Alternative is to add avoid immediate end of job (using sleep for example) et manually retrieve logs from docker container.

firecow:
closes #611


atli-c (approved):
LGTM 👍

PR #631, opened 2022-10-17
changes the single rsync for a more focused rsyncs

gbenguria:
changes the single rsync for a more focused rsyncs

using GIT_STRATEY GIT_SUBMODULE_STRATEGY and GIT_SUBMODULE_PATHS to improve start speed in large repos

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -114,7 +114,6 @@ export async function handler (args: any, writeStreams: WriteStreams): Promise<R
         await state.incrementPipelineIid(cwd, stateDir);
         const pipelineIid = await state.getPipelineIid(cwd, stateDir);
         parser = await Parser.create(argv, writeStreams, pipelineIid);
-        await Utils.rsyncTrackedFiles(cwd, stateDir, ".docker");
         await Commander.runPipeline(argv, parser, writeStreams);
         writeStreams.stderr(chalk`{grey pipeline finished} in {grey ${prettyHrtime(process.hrtime(time))}}\n`);
     }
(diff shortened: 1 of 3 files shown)


firecow:
I think can and should be bolied down to a single rsync system call, if possible.

gbenguria:
omg!! race condition!! 
I not 100% as I have not constrast this thoughts. If we asume that gitlab-ci-local uses the same runner, the jobs should run sequentially, but I am no sure of the order ... if it is based in the order of the jobs in th gitlab-ci.yaml then race conditions should not take place. If the order is random then yes we may find different results in some odd cases.
If we assume some kind of parallelization then that implys multiple runners, in that case we will need to control runners and store the data for each one separatelly. For the moment, I prefer the single runner approach.

firecow:
Jobs run in parallel on a single gitlab-runner instance, and they run in parallel in gitlab-ci-local as well.

firecow:
I would object against putting a mutex in front of every rsync command.

Have a look at how the shell isolation jobs build folders are created, and find a smart way of deducting build folders for docker jobs with simlar git strategies is better.

gbenguria:
same here, will be possible to put this as draft until I add the necessary tests? sorry for that, again.

gbenguria:
upsss I have find the convert to draft link :-)

gbenguria:
reading the shell runner documentation y have seen that the repo is cloned under a path that includes concurrent_id https://docs.gitlab.com/runner/executors/shell.html. That could minimize the possibility of race conditions. But implementing that also require extra disk space for the different copies of the repo.
Maybe a concurrency option can be added to control the number of jobs that can be run in parallel.

gbenguria:
it is replaced with pull request #647 that is build over #634

PR #652, opened 2022-11-08
Support artifacts:when:always

kouk:
Currently when artifacts:when is set to always the artifacts of a failed job aren't copied out.

Ref: https://docs.gitlab.com/ee/ci/yaml/#artifactswhen

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -268,7 +268,7 @@ export class Job {
         return this.jobData["description"] ?? "";
     }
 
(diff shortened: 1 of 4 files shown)


firecow (approved):
Epic stuff. You want a contributor invite @kouk ?

kouk:
@firecow my team's using your tool heavily, so yeah, I could help out for sure. Although I must tell you I am no typescript developer. In fact this is the very first time I've written a line of TS. Kind of winging it with my knowledge of JS.

PR #657, opened 2022-11-09
Pull image if podman fails image inspect

### LINKED ISSUE #655: "podman image inspect" fails with a different error message than "docker image inspect"
### Issue
Podman is creating a different error message than docker when `... image inspect docker.io/tmknom/markdownlint:0.32.2` fails.

Based on this fact, the following error catch is not working and does break the whole execution of `gitlab-ci-local` since it can't find the phrase from the following code line:

https://github.com/firecow/gitlab-ci-local/blob/07fbc95aca92ad4c437cc8a789861ca46ccdac5c/src/job.ts#L692

### Dockers Output

```
➜  ~ docker image inspect docker.io/tmknom/markdownlint:0.32.2
[]
Error: No such image: docker.io/tmknom/markdownlint:0.32.2
```

### Podmans Output

```
➜  ~ podman image inspect docker.io/tmknom/markdownlint:0.32.2
[]
Error: inspecting object: unable to inspect "docker.io/tmknom/markdownlint:0.32.2": failed to find image docker.io/tmknom/markdownlint:0.32.2: docker.io/tmknom/markdownlint:0.32.2: image not known
```

### Suggested Solution

1. Make the error handling more "general". So we don't care with what error the `docker image inspect` fails but we want to pull anyways.
2. Create another if case where it also checks for the podmans output e.g. `image not known`.
3. Check if the docker image exists in a different way like this: https://stackoverflow.com/a/30543453
4. Completely ditch the check and the pull since docker pulls on a `run` anyways when the image does not exist

Are there any opinions to that?

firecow:
closes #655

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -690,7 +690,7 @@ export class Job {
         try {
             await Utils.spawn(["docker", "image", "inspect", imageToPull]);
         } catch (e: any) {
-            if (e.stderr?.includes(`No such image: ${imageToPull}`)) {
+            if (e.stderr?.includes("No such image") || e.stderr?.includes("failed to find image")) {
                 await Utils.spawn(["docker", "pull", imageToPull]);
                 const endTime = process.hrtime(time);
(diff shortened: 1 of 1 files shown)


atli-c (approved):
LGTM

PR #690, opened 2022-12-10
fix(689): Adding case to execute pipelines with absent optional jobs

### LINKED ISSUE #689: Job declared as optional raises pipeline error if not exists due to rules
```yml
skipped-job:
  stage: build
  when: always
  script:
    - echo "Job skipped!"
  rules:
    - when: never

executed-job:
  stage: test
  script:
    - echo "Job executed!"
  needs:
    - job: skipped-job
      optional: true
```

**Expected behavior**
The job `executed-job` should be executed when the `skipped-job` was not added to the pipeline.
(tested same functionality in a gitlab pipeline and it works)

**Assumption**
It might be, that gitlab-ci-local is analyzing the needs before the rules.

alexander-matthiesen:
closes #689

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -95,6 +95,9 @@ export class Executor {
                 if (j.when === "never" && !need.optional) {
                     throw new AssertionError({message: chalk`{blueBright ${j.name}} is when:never, but its needed by {blueBright ${job.name}}`});
                 }
+                if (j.when === "never" && need.optional) {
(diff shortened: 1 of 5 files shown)


alexander-matthiesen:
Some tests as well in our project showed, that the issue is resolved with this MR.

firecow (approved):
Good stuff 😁

PR #694, opened 2022-12-17
feat: list option for csv

alexander-matthiesen:
This MR is adding the functionality to output the pipeline jobs in CSV format which enables easier testing scenarios since the "pretty" format is not really predictable with the indentations after the pipeline changes.


alexander-matthiesen:
@firecow , I wasn't sure about adding the `--list-json-all` functionality to this MR since the all functionality is already in `--list-json` and it would break that functionality if I would streamline it to `--list` and `list-all`.

I can add this if wanted.

Also I had to raise the timeout for the dind test since 30s was not enough (see previous pipelines).

firecow:
@alexander-matthiesen 

Is the semi colon detection necessary?

If you wrap the values in quotes or double quotes, you will still have functioning CSV.

firecow:
> @firecow, the example was committed on purpose because I was lacking a general example for the documentation screenshots. I can remove it again if you want.
> 
> Should the description be wrapped in quotes only when there is a semicolon or always? 

Lets have the examples folder contain "real life" examples.

Just quote em always, if you ask me.

alexander-matthiesen:
> What documentation screenshots?

I don't mean screenshots but the code examples in the README.md. I was struggling a bit figuring out on which basis they were created to have a uniform documentation but this might be a little to much.

I applied your suggestions.

firecow (approved):
Epic sauce.

PR #706, opened 2022-12-30
feat: Adding --umask (default true) option

### LINKED ISSUE #671: --disable-umask cli option
gitlab-ci-local seems to always do `docker create -u 0:0 ..`

This causes problems with file ownership and running applications/testing

`docker create ...` without -u

**Additional context**

alexander-matthiesen:
closes #671

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -92,6 +92,10 @@ export class Argv {
         return this.map.get("cleanup") ?? true;
     }
 
+    get umask (): boolean {
+        return this.map.get("umask") ?? true;
+    }
+
     get privileged (): boolean {
         return this.map.get("privileged") ?? false;
     }
--- a/src/index.ts
+++ b/src/index.ts
@@ -135,6 +135,11 @@ import {AssertionError} from "assert";
             description: "Enable docker mount based caching",
             requiresArg: false,
         })
+        .option("umask", {
+            type: "boolean",
+            description: "Sets docker user to 0:0",
+            requiresArg: false,
+        })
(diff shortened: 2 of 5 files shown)


alexander-matthiesen:
> We cannot make this option default to true, when gitlab's default is false.

So currently the `--umask` activates the `--user 0:0` and `--umask` is set to true by default. 
Do I have a logic issue in my head or do I misunderstand the `umask` setting?

If we set `--umask` default to `false`, the `--user 0:0` would not be added and this would be a breaking change to the current setup.

firecow:
> > We cannot make this option default to true, when gitlab's default is false.
> 
> So currently the `--umask` activates the `--user 0:0` and `--umask` is set to true by default. 
> Do I have a logic issue in my head or do I misunderstand the `umask` setting?
> 
> If we set `--umask` default to `false`, the `--user 0:0` would not be added and this would be a breaking change to the current setup.

Let me look at the code again... Maybe I got it wrong.

firecow:
> > > We cannot make this option default to true, when gitlab's default is false.
> > 
> > So currently the `--umask` activates the `--user 0:0` and `--umask` is set to true by default. 
> > Do I have a logic issue in my head or do I misunderstand the `umask` setting?
> > 
> > If we set `--umask` default to `false`, the `--user 0:0` would not be added and this would be a breaking change to the current setup.
> 
> Let me look at the code again... Maybe I got it wrong.

Yes i did. It has to be true of course 🤟

firecow (approved):
lgtm

PR #716, opened 2023-01-04
Env specific file path vars

### LINKED ISSUE #715: Environment specific variables from file paths

Currently, while a cross-environment variable can be sourced from a path, e.g. in `.gitlab-ci-local-variables.yml` the below would allow using the kubernetes CLI in the job:
```
KUBECONFIG: ~/.kube/config
```
the same feature cannot be used with environment specific values, e.g.:
```
---
KUBECONFIG:
   values:
       '*devel*': ~/.kube/config.devel
       '*staging*': ~/.kube/config.staging
       '*production*': ~/.kube/config.production
```
Adding `type: file` doesn't fix the issue, since it will just create a file with the path as the contents.

Be able to use env specific variables with content from file paths.

alternatively one could embed their files as YAML multiline literals.

**Additional context**

none

kouk:
An attempt to resolve #715. In this implementation, the absence of a `type` field for the variable definition will result in values automatically being treated as file paths or as literal values.
I've added a failing test case.
I'm not very happy with this solution, as it's kind of ambiguous. Also you can mix and match, with the variables for some environments coming from files and for others from literal values (this may be an advantage).

An alternative would be to define a new constant for the `type` field, such as `variable-from-file`, that enables the feature.


firecow:
Outdated, closing for now.

PR #723, opened 2023-01-14
Replace apt-key command for Debian repository setup

detly:
`apt-key` has been deprecated since 2019 (I think, I was unable to find the original notice). It will probably stop working in Ubuntu post-2023. The newer alternatives are to

1. [Organise keys separately]() somewhere *not* `/etc/apt/trusted.gpg.d/` and use the `[signed-by=...]` key in the sources line
2. (Preferred) use the Deb822 format with the GPG key in the file

This PR addresses (1) by updating the `.list` file and the instructions; and (2) by providing a Deb822 format `.sources` file and instructions for installing that. Note for your own workflow the contents of `ppa/pubkey.gpg` are now duplicated in `ppa/gitlab-ci-local.sources`.

I've tested both of these methods on Ubuntu 22.10 but by copying locally, not from the URL in the README.

See:

- https://wiki.debian.org/DebianRepository/UseThirdParty
- https://repolib.readthedocs.io/en/latest/deb822-format.html#deb822-format
- https://unix.stackexchange.com/a/699356
- https://askubuntu.com/a/1307181


firecow (approved):
Looks good to me

PR #732, opened 2023-01-23
Tolerate needs.pipeline

kouk:
We currently cannot test a pipeline configuration that is part of a multi-project or parent-child pipeline configuration if a job has a `needs:pipeline` field, even if the dependency can be mocked somehow (e.g. passing some upstream variable on the CLI). The error is something like:
```
needs: [my-needed-job] for my-job could not be found
```

When a job has needs.pipeline, because it depends in some way on a job from another pipeline we currently fail to validate the pipeline because the needed job isn't in the current list of jobs of the pipeline. However instead of failing, it is more useful to ignore the particular dependency and emit a warning.

An alternative would be to continue failing and force the user to manually edit the file to comment out the entry in `needs:pipeline` entry.

**Additional context**

I made a choice to include the pipeline attribute in the `Job.needs` data structure, which would indicate to the validator and executor that the specific dependency should be ignored. I also updated the Validator to generate a list of warnings that can be output all together. Alternatives I considered were a) updating the Job.needs method to just skip the entries of Job.needs that contained the pipeline field because that method isn't memoized so I couldn't emit a warning there (if you would prefer we memoize it then this alternative would be a good one), and b) passing the `writeStreams` object to the Validator so that it could output the warnings itself, but I feel the parser might be a better place to control the "interaction" with the user.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@
 !/docker-npm
 
 /.gitlab-ci-local/
+
+.DS_Store
+
+.vscode
--- a/src/job.ts
+++ b/src/job.ts
@@ -188,15 +188,16 @@ export class Job {
         return Utils.safeDockerString(this.name);
     }
 
-    get needs (): {job: string; artifacts: boolean; optional: boolean}[] | null {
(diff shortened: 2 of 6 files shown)


firecow (approved):
Looks good to me...

PR #815, opened 2023-04-09
Set yargs configuration greed-arrays to false

### LINKED ISSUE #810: --volume should work before the job name
Hello 👋

for now, `--volume` has to be after a named job, else the job is ignored and all is launched :

Screenshot 2023-04-05 at 12 23 58

So I have to write it this way for now : `gitlab-ci-local 📦✅-auth-server-package-and-test --volume my-volume:/host/`

It's then harder to change only the job name when launching multiple times. And I can't easily make an alias for a permanent volume mount, like `--volume /var/run/docker.sock:/var/run/docker.sock`

Or I missed a configuration to set this locally in a project ?

firecow:
closes #810

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -16,6 +16,7 @@ import {AssertionError} from "assert";
 (() => {
     const packageJson = JSON.parse(fs.readFileSync(path.join(__dirname, "../package.json"), "utf8"));
     yargs(process.argv.slice(2))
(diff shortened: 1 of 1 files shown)


bcouetil (approved):
I understood from documentation that --volumes was, by default, greedy, and "swallowing" more than one word.

Approved !

PR #832, opened 2023-04-25
Add ulimit configuration option to env file

alex116:
I need the ulimit option for `docker run` because I don't want to change the ulimit of the whole system, just the docker images that I run. This commit would enable using this option in the env file to modify the execution of `docker` to contain the `--ulimit nofile=8000:16000` parameter/values.   

Currently, only the values can be modified in the env file like so:
`ULIMIT=8000:16000`  
If this is too specific, I could make it more general - so that the env file can be configured to contain 
`ULIMIT="nofile=8000:16000"`  
Or something similar so that the nofile or nprocs or whatever other option can be set by the user.


firecow (approved):
LGTM

PR #854, opened 2023-05-10
#641: allow executing single matrix job

### LINKED ISSUE #641: Option to execute single parallel matrix job
```yml
---
job:
  script:
    - echo "$var"
  parallel:
    matrix:
      var: [a,b,c]
```

this leads to:

```bash
gitlab-ci-local --list

parsing and downloads finished in 27 ms
name     description  stage   when        allow_failure  needs
job [a]               test    on_success  false
job [b]               test    on_success  false
job [c]               test    on_success  false
```

```bash
gitlab-ci-local "job [a]"

parsing and downloads finished in 26 ms
job [a] could not be found
```

**Expected behavior**

job `job` is actually run for instance `[a]`.

**Note**: Everything works when you remove the `[a]`, however then all jobs of the matrix are triggered in parallel which messes up your terminal quite a bit and is not the intended behaviour in our use case. (we build multiple architectures in parallell; for local development it would make sense to only build the locally testable one)

**Host information**
MacOS Ventura 13.0 (22A380)
gitlab-ci-local 4.34.0

**Additional context**

hverlin:
Resolves #641


hverlin:
Might conflict with #845

hverlin:
I think it's an edge case and I would not worry too much about it 😅 

On gitlab.com, I get the following behavior. Note that the job name has a `:` for matrix job now?
image
image
Seems that if someone creates the same job name as the matrix job, then we see only one of them in the UI

firecow:
There is a inconvenience if we decide to roll with your suggestion.

```
test-job [beb, stage]
test-job [stage, beb]
```
Only one will execute the job the developer is after.
 
 ```
 test-job --matrix="NAME=beb,TIER=stage"
 test-job --matrix="TIER=stage,NAME=beb"
 ```
Here the developer is forced to know about the matrix variables, and ordering is irrelevant for him.

firecow:
image

```
---
test-job:
  parallel:
    matrix:
      - SITE: denmark
        TIER: stage
      - SITE: sweden
        TIER: stage
  script: echo "Ok"

"test-job: [denmark, stage]":
  script: echo "Over"
```

Ok, you can actually override a parallel.matrix job :smile: I didn't know that, lol...

hverlin:
Seems like unspecified behavior on GitLab side 😅 

Do you think it's ok to roll-it as is? I think it would be convenient to leave it working as implemented in this MR so that you can execute jobs as they appear in the output of 
```
gitlab-ci-local --list
```

I feel like having the matrix option would be harder to discover. Even if we add it, we would need to add a message to let the users know about it if the command line argument is matching one of the matrix jobs or try to run it? 
Also, I think the `matrix` option would need to be specified if one tries to execute multiple matrix jobs?

firecow:
Implementing it "as-is" would be ok.
If we resolve https://github.com/firecow/gitlab-ci-local/issues/860 before next release

firecow:
@hverlin The colon just conflicted here. 
Let me know, when you've resolved, and we will add this to master.

firecow (approved):
Let's move to master

PR #865, opened 2023-05-16
Parallel service start

repomaa:
Starts services in parallel. Significant speed up with many services or long health checks.


src/job.ts
```diff
@@ -1086,6 +1089,8 @@ export class Job {
             aliases.push(serviceAlias);
         }
 
+        const uniqueAlias = aliases[aliases.length - 1];
```
repomaa, line 1092: if you have two services using the same image in the job, the health check will just be checking either one. in those cases it's likely that there are unique alias definitions for the services which we can try to use here.

firecow: Hmm... I wonder how gitlab.com actually pulls this one of.

firecow (approved):
Changes looks good to me

PR #867, opened 2023-05-16
Fix failing parallel-matrix test

firecow:


hverlin (approved):

hverlin:
@firecow there was still a small issue due to the `:` in the job name now. I have fixed it in #854 as well as added a test to make sure the jobs run successfully

PR #889, opened 2023-05-29
Add "build" alias to main container if services are enabled (with tests)

bbrala:
By default in gitlab the main job has an alias "build", added build alias to container when services are used.

See https://gitlab.com/gitlab-org/gitlab-runner/-/issues/27060

This enables connecting to the main container using the build alias for example for chromedriver by using "build" as the hostname.

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -635,7 +635,7 @@ export class Job {
 
             if (this.services?.length) {
                 dockerCmd += `--network gitlab-ci-local-${this.jobId} `;
-                dockerCmd += `--network-alias=build `;
+                dockerCmd += "--network-alias=build ";
             }
 
             dockerCmd += `--volume ${buildVolumeName}:/gcl-builds `;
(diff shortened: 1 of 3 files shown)


firecow:
@bbrala Is this feature actually present on gitlab.com at the moment?

bbrala:
https://gitlab.com/gitlab-org/gitlab-runner/-/blob/main/executors/docker/docker.go#L537

yeah it is in the source.

bbrala:
See this comment: https://gitlab.com/gitlab-org/gitlab-runner/-/issues/27060#note_635488187

bbrala:
Only thing I'm not sure of, should this alias also be added if no services. It seems the gitlab source might actually do that, but also seems redundant.

Also the test could be better to test if no service is added the alias is not present. But that is your call if that needs adding imo.

bbrala:
That is a weird failure :/

firecow (approved):
Looks good to me

bbrala:
Great! Thanks. 

Any changes you can send a message here when you release a new version?

bbrala:
Never mind the comment, I subscribed to releases :)

PR #905, opened 2023-06-16
Allow single references in services field

### LINKED ISSUE #904: Single reference (i.e. no array) in services field results in a TypeError
```yml
---
stages:
  - minimal_example

.something-to-ref:
  services:
    - name: "alpine:latest"

minimal_example:
  stage: minimal_example
  services: !reference [.something-to-ref, services]
  script: echo "hello world"
```

**Expected behavior**
https://gitlab.com/jameshi16/gitlab-ci-local-services-issue/-/jobs/4485354756 (green pipeline)

There are two problems here:
- https://github.com/firecow/gitlab-ci-local/blob/master/src/data-expander.ts#L151 reassigns `jobData.services = services`, which basically invalidates the line before.
- If a reference is passed directly without being in a YAML array, an object, like `{ referenceData: [ '.something-to-ref', 'services'] }` is given to the `reference` function call, which doesn't get parsed right, and also results in a `TypeError` (`jobData.services.flat` is not a function)

**Host information**
Ubuntu
gitlab-ci-local 4.1.33

**Additional context**
I'll open a PR shortly in an attempt to fix the issues above, and link it to this issue.

jameshi16:
Addresses https://github.com/firecow/gitlab-ci-local/issues/904, where `gitlab-ci-local` fails to correctly parse GitLab CI configuration if there is a singular `!reference` in the services field.

closes #904

### DIFF
--- a/src/data-expander.ts
+++ b/src/data-expander.ts
@@ -146,9 +146,9 @@ export function servicesEach (jobName: string, gitlabData: any) {
     const services = jobData.services;
     if (!services) return;
 
+    jobData.services = Array.isArray(services) ? services : [services];
     reference(gitlabData, jobData.services);
     jobData.services = jobData.services.flat(5);
-    jobData.services = services;
 
     for (const [i, s] of Object.entries<any>(jobData.services)) {
         jobData.services[i] = servicesComplex(s);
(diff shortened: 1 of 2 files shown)


firecow (commented):
Can you add a little test to ensure we don't break this in the future?

firecow (approved):
Looks good to me.
Thanks for contributing @jameshi16 :+1:

Issue #703, opened 2022-12-22
Limit concurrent jobs


olkitu:
Add possible limit concurrency jobs. I have some tests example run normally over 10 jobs same time but no way run in local computer so many concurrent jobs.

bcouetil:
I realize that I'm doing a lot of unnecessary work to get around this limitation on my biggest projects.

A big +1 for this 👍

firecow:
This is just for preventing your PC's getting too hot right?
Is there anything buggy going on, that we should fix?

bcouetil:
CPU is not the main concern.

It could allow too much RAM, which makes jobs, or even the Docker system, crash.

olkitu:
Yes too much load when example have over 10 jobs in same stage. Normally on CI have scalable environment to run on multiple nodes.

naweiss:
Opened a PR for this

bcouetil:
Thank you all, nice feature 🙏

PR #928, opened 2023-07-08
Add an option to limit Executor concurrency

naweiss:
Closes issue #703

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -13,7 +13,7 @@ export class Executor {
             startCandidates = Executor.getStartCandidates(jobs, stages, potentialStarters, argv.manual);
             if (startCandidates.length > 0) {
                 await PromisePool
-                    .withConcurrency(argv.jobs ?? startCandidates.length)
+                    .withConcurrency(argv.concurrency ?? startCandidates.length)
                     .for(startCandidates)
                     .process(async (job, index, pool) => {
                         return job.start();
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -2,7 +2,7 @@ import chalk from "chalk";
 import {Job} from "./job";
 import assert, {AssertionError} from "assert";
 import {Argv} from "./argv";
-import { PromisePool } from '@supercharge/promise-pool'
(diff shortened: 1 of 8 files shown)


naweiss:
Originally I thought about solving this problem using a simple slicing of the jobs array. But, this has the disadvantage of having to wait for a long job until running the next batch of jobs.
So, ultimately I used `PromisePool` which generate the next job immediately when the previous ended.

firecow:
Good idea to use a concurrency library.

Add a test and we are good to go 👍

naweiss:
I had hard time writing a test that jobs runs sequentially instead of parallel. I ended up using the `__snapshot__` method used in the padding tests. hope it is fine

firecow:
Yeah, its fine.

Let us rename the option to --concurrency or --concurrent 👍

naweiss:
Do you know how I solve the following linting warning?
```
Warning:   18:42  warning  '_' is defined but never used    @typescript-eslint/no-unused-vars
```

naweiss:
I have to fix some minor thing before the merge

naweiss:
There was something messed up with the author of the commits. It is fixed now

src/executor.ts
```diff
@@ -2,19 +2,23 @@ import chalk from "chalk";
...
         let startCandidates = [];
 
         do {
-            const promises = [];
             startCandidates = Executor.getStartCandidates(jobs, stages, potentialStarters, argv.manual);
-            for (const s of startCandidates) {
-                promises.push(s.start());
+            if (startCandidates.length > 0) {
+                await PromisePool
+                    .withConcurrency(argv.concurrency ?? startCandidates.length)
```
hverlin, line 16: suggestion: use `os.cpus().length - 1` as the default value? Pass `-1` for unlimited concurrency?

firecow: If developers have problems with concurrency they can explicitly set it.

People have been running with unlimited up until now, no need to change that default.

naweiss:
Anything else I need to fix?

firecow (approved):
Nope all looks good, merging

PR #939, opened 2023-07-19
Bugfix/parallel matrix artifacts copy

### LINKED ISSUE #923: parallel matrix artifacts are not copied
Using the following .gitlab-ci.yml file:
```yml
stages:
    - prebuild
    - build

setup:
    image: ubuntu:18.04
    stage: prebuild
    parallel:
        matrix:
            - FILE_NAME: [aaaa, bbbb]
    script:
        - touch "${FILE_NAME}"
    artifacts:
        paths:
            - ./${FILE_NAME}

build_package:
    image: ubuntu:18.04
    stage: build
    script:
        - cat aaaa
        - cat bbbb
```

You get the following error:
```
build_package > cat: aaaa: No such file or directory
```

**Expected behavior**
The expected behaviur is that the build_package job will finish successfully like [here](https://gitlab.com/naweiss/gitlab-ci-local-testing/-/jobs/4614933576)

**Host information**
Ubuntu 20.04
gitlab-ci-local 4.41.2

**Additional context**
It is important to notice that if you add to build_package:
```yml
build_package:
    needs: 
        - setup
```
gitlabi-ci-local print the correct output.

And, if you does not use parallel matrix for setup like so:
```yml
setup:
   ...
    variables:
        FILE_NAME: aaaa
...
build_package:
    ...
    script:
        - cat aaaa
```
gitlabi-ci-local also print the correct output.

naweiss:
closes #923

### DIFF
--- a/src/producers.ts
+++ b/src/producers.ts
@@ -29,7 +29,7 @@ export class Producers {
         for (const potential of jobs) {
             if (potential.artifacts == null) continue;
             if (potential.when == "never") continue;
-            if (!producerSet.has(potential.baseName)) continue;
+            if (!producerSet.has(potential.name) && !producerSet.has(potential.baseName)) continue;
             producers.push(potential);
         }
         return producers.map(producer => {
--- a/tests/test-cases/artifacts-sequence/.gitignore
+++ b/tests/test-cases/artifacts-sequence/.gitignore
@@ -1,0 +1,4 @@
+/file1
+/file2
+/file3
+/file4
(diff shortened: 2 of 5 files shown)


naweiss:
closes #923

naweiss:
To give some context, producer is the one who check if artifact should be copied:
https://github.com/firecow/gitlab-ci-local/blob/90e5f2deea3dbd60c9bfc1a27ca59062a6c923de/src/producers.ts#L6-L38

There are two main cases:
- one of `job.needs` or `job.dependencies` is present, in which case they will contain the `baseName` of the needed job
- non of them is present, in all of the previous job names will be considered as needed

firecow (approved):
Looks good to me 🤟
I've sent you an invite @naweiss , so you don't have to "fork around" 🤣

PR #946, opened 2023-07-30
Use p-map instead of promise-pool

firecow:
@naweiss Have a look. See https://github.com/firecow/gitlab-ci-local/discussions/943 for more information.

promise-pool package doesn't throw on promise rejections, and we want that to happen, so developers can see error messages.

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -2,7 +2,7 @@ import chalk from "chalk";
 import {Job} from "./job";
 import assert, {AssertionError} from "assert";
 import {Argv} from "./argv";
-import {PromisePool} from "@supercharge/promise-pool";
+import pMap from "p-map";
 
 export class Executor {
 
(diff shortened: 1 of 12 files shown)


naweiss:
@firecow can you give a minimal example to check the issue against?

firecow:
> @firecow can you give a minimal example to check the issue against?

Good idea, we don't want this to come back 👍

firecow:
@naweiss I've added a test to prevent regression :+1:

src/executor.ts
```diff
@@ -12,12 +12,8 @@ export class Executor {
         do {
             startCandidates = Executor.getStartCandidates(jobs, stages, potentialStarters, argv.manual);
             if (startCandidates.length > 0) {
-                await PromisePool
-                    .withConcurrency(argv.concurrency ?? startCandidates.length)
-                    .for(startCandidates)
```
naweiss, line 17: I looked into the documentation of "@supercharge/promise-pool", and does seems to support throwing on promise rejections by adding:
```
.handleError(async (error) => { throw error; })
```
So, we should decide which library will be better.

firecow: p-map is made by https://github.com/sindresorhus and has zero dependencies, it's fine 👍

naweiss (approved):

PR #953, opened 2023-08-05
Add flags negation to help

### LINKED ISSUE #924: Add no-artifacts-to-source to help
One of command line options of gitlab-ci-local is `--artifacts-to-source` which hints that gitlab-ci-local does not copy the artifacts to the source directory by default which is not true. also, from looking into the output of `--help` it is unclear how to disable this option.
I accidently discovered that it can be disabled by `--no-artifacts-to-source` after looking at the README of this repo. But, I don't think this is a good way to document this option. A batter way would be to add it to the output of `--help`.

**other alternatives**
1. Making this option false by default which will make more sense
2. Add a message to help clearifaying you can add a **no** prefix to boolean options

naweiss:
Closes #924

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -54,7 +54,7 @@ process.on("SIGINT", (_: string, code: number) => {
             command: "$0 [job..]",
             describe: "Runs the entire pipeline or job's",
         })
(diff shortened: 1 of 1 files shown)


firecow (approved):
Looks good 👍

PR #996, opened 2023-10-02
Add --container-executable option. Fix #922.

### LINKED ISSUE #922: allow configuration of "containerd" executable
I am using podman, which features a mostly compatible CLI for docker commands.
It would be great if I could use this setup automatically

I would like this tool to test both `podman` and `docker` and use the
tool that was found (and don't even try to launch anything if nothing was found).

Allowing the tool path to be used would be another option

Well, a simple shell-script wrapper works of course.

ebousse:
Fix #922

Add a new option `--container-executable` which can be used to choose any executable in place of  `docker` (which remains the default value).

I haven't added any test cases so far, for two reasons (1) I could not find any test cases about checking that args were processed correctly, so adding one seemed out of place? (2) I did not want the test environment to required both Docker and Podman?

But maybe I could write a test case that sets a gibberish executable, and check that it fails as expected?


src/job.ts
```diff
@@ -1102,7 +1102,7 @@ export class Job {
             await Promise.any(Object.keys(imageInspect[0].Config.ExposedPorts).map((port) => {
                 if (!port.endsWith("/tcp")) return;
                 const portNum = parseInt(port.replace("/tcp", ""));
-                const spawnCmd = ["docker", "run", "--rm", `--name=gcl-wait-for-it-${this.jobId}-${serviceIndex}-${portNum}`, "--network", `gitlab-ci-local-${this.jobId}`, "willwill/wait-for-it", `${uniqueAlias}:${portNum}`, "-t", "30"];
+                const spawnCmd = [this.argv.containerExecutable, "run", "--rm", `--name=gcl-wait-for-it-${this.jobId}-${serviceIndex}-${portNum}`, "--network", `gitlab-ci-local-${this.jobId}`, "willwill/wait-for-it", `${uniqueAlias}:${portNum}`, "-t", "30"];
```
skuethe, line 1105:
```suggestion
                const spawnCmd = [this.argv.containerExecutable, "run", "--rm", `--name=gcl-wait-for-it-${this.jobId}-${serviceIndex}-${portNum}`, "--network", `gitlab-ci-local-${this.jobId}`, "docker.io/willwill/wait-for-it", `${uniqueAlias}:${portNum}`, "-t", "30"];
```

We need to adapt this to use the complete docker image URL here. Otherwise it will break:

```
[...]
[13:40:52    21 s] tests started service image: docker.io/some/image:tag with aliases: docker.io-some-image, docker.io__some__image, myalias in 384 ms
[13:40:52    21 s] tests service image: docker.io/some/image:tag healthcheck failed
[13:40:52    21 s] tests   Command failed with exit code 125: docker run --rm --name=gcl-wait-for-it-967338-0-4444 --network gitlab-ci-local-967338 willwill/wait-for-it myalias:4444 -t 30
[13:40:52    21 s] tests   Error: short-name resolution enforced but cannot prompt without a TTY
[...]
```

ebousse: Good point, some `podman` setups don't have `docker.io` as default registry.

In fact it would make sense to add `docker.io/` to all image references in the repository, including examples and test cases, to make it explicit that all images are to be found in the Docker Hub.

I'll prepare extra changes for that.

skuethe:
Just one small suggestion from myself:
I would prefer to use the official URLs for "official images" of `docker.io/library/[image]`.  
See:  
https://docs.docker.com/registry/introduction/#understanding-image-naming

I do think the short handle (`docker.io/[image]`) would work within podman, but as far as I know that is a convenience feature from podman and it just translates it to the official URL mentioned above.

ebousse:
> Just one small suggestion from myself: I would prefer to use the official URLs for "official images" of `docker.io/library/[image]`. See: https://docs.docker.com/registry/introduction/#understanding-image-naming
> 
> I do think the short handle (`docker.io/[image]`) would work within podman, but as far as I know that is a convenience feature from podman and it just translates it to the official URL mentioned above.

AFAIK `docker.io/image` il not really a short handle, it's simply that all Docker Hub's "_Official Images_" simply do not have  a group/library, therefore you have no choice but to write the URL in that way.

For example, the [alpine image](https://hub.docker.com/_/alpine) can only be obtained with:

```
podman container pull docker.io/alpine
```

and cannot be obtained with:

```
podman container pull docker.io/_/alpine
```

which throws an `Error: invalid reference format`.


Or maybe I misunderstood what you were talking about?

skuethe:
> @skuethe It seems unfortunately you did not see my edit, where I completely removed my initial message after I understood all this. Thanks for writing the explanations though!

oh, sorry. That was overlapping :)

ebousse:
Pushed the following changes:

- I had mistakenly committed a wrong search and replace with wrong indents in `image:` declarations, now fixed
- Added a `npm run` command called `test-except-dind` to run all test suites _except_ the one testing the use of the Docker `dind` image. This allows running only test cases that should pass with Podman, since `dind` cannot work with Podman AFAIK.
- As suggested by @skuethe, added `library` as an explicit prefix to all docker hub images that are not in a group

I think I should change the PR title to reflect the changes on the test suite as well?

And overall, with the latest changes, I feel more confident with the branch. More feedback welcome.

ebousse:
As the CI showed:
- I missed a YAML linting issue (btw there is no YAML linter configured on the `package.json`?), now fixed
- I created a problem in the dind test case due to referencing the image as `docker.io/library/docker:dind` instead of `docker:dind` previously. I've added a `docker` alias to fix the problem.

src/index.ts
```diff
@@ -225,6 +225,11 @@ process.on("SIGINT", (_: string, code: number) => {
             description: "Limit the number of jobs that run simultaneously",
             requiresArg: false,
         })
+        .option("container-executable", {
+            type: "string",
+            description: "Command to start the container engine (Docker or Podman)",
```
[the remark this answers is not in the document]
firecow: @ebousse Can you put "docker" or "podman" instead of first lette uppercase?

firecow (approved):
Looks good to me :+1: Good job

PR #1016, opened 2023-10-30
Change hard-coded docker commands to respect --container-executable

ssiegel:
The original implementation of `--container-executable` in #996 missed these instances.


firecow (approved):
Looks good to me

PR #1017, opened 2023-10-30
Try to auto-detect docker or podman binary

ssiegel:


firecow:
Increasing upstart time for the majority of developers is a bad idea i feel..

Let's add a podman section to the readme instead, explaining how to use --container-executable, when you wanna use podman.

ssiegel:
I totally see the point, and I agree that adding the autodetection is indeed quite heavy to add to every start, when it might not be needed, e.g. when just using the shell executor or autocompletion.

My main issue is that without a global configuration file, it is inconvenient to always have to add the `--container-executable` option. To define a shell alias for that doesn't look like the correct solution to me. I realize now that I can set the `GCL_CONTAINER_EXECUTABLE` environment variable, which is better, but maybe there is an automatic way without the startup impact for the most common case:

I updated the pull request to run the container executable autodetection only if we need it. Then we're going to run `docker` multiple times anyway (17 `docker` invocations in the following minimal hello world `.gitlab-ci.yml`). Adding a single `docker version` to that looks reasonable to me.

```yaml
hello-world:
  image: alpine:latest
  script:
    - echo "hello world"
```

ssiegel:
I'm sorry, I totally missed the fact that it's possible to put global configuration into `~/.gitlab-ci-local/.env`, which completely solves my initial problem. So feel free to either just close this pull request, or use it if you agree that it might be useful for gitlab-ci-local to just work out of the box with just podman installed, without docker symlink.

firecow:
I don't agree that all docker executor jobs should be have increased upstart time, when its so "easy" to add a `export GCL_CONTAINER_EXECUTABLE=pod` in `.bashrc` or use the `~/.gitlab-ci-local/.env` :smile: 

Closing, but thanks for the PR's good work.

Issue #371, opened 2022-01-26
gitlab-ci-local --fetch-includes should handle imports like gitlab (http/https as remote import)


typoworx-de:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
include:
  - 'https://gitlab.com/..../raw/master/pipeline/my-deploy.yml'
```

**Expected behavior**
Gitlab will fetch the "remote" Yaml-Config. gitlab-ci-local will fail even with ```gitlab-ci-local --fetch-includes```

firecow:
@typoworx-de Hmmm... Thought that already worked. I'll have a look.

firecow:
@typoworx-de Please paste a concrete example instead of the pseudo one.
We already have a test to verify that this functionality works in https://github.com/firecow/gitlab-ci-local/blob/master/tests/test-cases/include-remote/.gitlab-ci.yml

So we need more info to debug from.

Reopen the issue, once you got a real life example, that I can debug.

firecow:
Oh, i think I know, what the issue is now...
http|https has specielt mening, and auto interprets as `remote:` on it

firecow:
@typoworx-de Fixed in https://github.com/firecow/gitlab-ci-local/commit/4db7812418334d90073340c3e46a7284507ccb63
Thanks for the bug report, even though I didn't get the problem at first :smile:

typoworx-de:
Hello @firecow,
thanks for resolving. Sorry I haven't been in office for the last few days, so I couldn't react to your questions. But I noticed you already figured out the "remote:" thing as Gitlab automatically determines http|https as remote without explicitly giving it the hint "remote:". Sorry if I have not been concrete enough. I'll try my best next time :-D

firecow:
@typoworx-de No problemo :smiley:

typoworx-de:
Great work, just checked the new release and it works like a charm now :-D

firecow:
Great, keep coming 👍

justinTM:
@firecow first of all thanks for making this. what an undertaking of a project. 

second of, i'm not sure if this is the right place, but i'm getting an include fetch error:

```console
% gitlab-ci-local "Plan Dev" --fetch-includes
Remote include could not be fetched https://gitlab.com/gitlab-org/gitlab/-/raw/HEAD/lib/gitlab/ci/templates/Workflows/Branch-Pipelines.gitlab-ci.yml
AggregateError
```

but if i plug in that URL to a browser, i get:

[...]

PR #1027, opened 2023-11-09
Print AggregateError's properly

firecow:
This PR will make it more clear what's going on in situations like this https://github.com/firecow/gitlab-ci-local/issues/371#issuecomment-1800126554

I can't really unittest it properly at the moment, so I'll leave screenshots on how I tested it here.

Added this to `job.ts` `execScripts` function

image

Ran `node src/index.js --cwd <path-to-whatever>` and got the following output.

image


FnuGk (approved):
LGTM

Issue #1036, opened 2023-11-15
Using --umask false causes permission errors on /gcl-builds


hatvik:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
job:
  script:
    - ls -l cmake/all_deps.cmake
    # Or
    - ci/my_script.sh
```
**Examples of failures**
`longevity                       > /gcl-cmd: line 40: ci/start_etcd.sh: Permission denied`

`longevity                       > ls: cannot access 'cmake/all_dependencies.cmake': Permission denied`

**Expected behavior**
You should be able to access files or execute scripts

**Containerd binary**
docker

**Additional context**
The problem is that the permissions on /gcl-builds are set to 0:0 and a+rw. And I am trying to run as a our image default, non-root ci user.

`drwxrwxrw-  2 root root  4096 Nov 14 15:14 cmake`

This means I cant list stuff in that directory, nor execute any files

firecow:
You need to `chmod +x ci/my_script.sh` or a regular remote Gitlab runner won't allow you to execute it either.

PR #1037, opened 2023-11-15
Set permissions for /gcl-builds to a+rwx

hatvik:
Fixes https://github.com/firecow/gitlab-ci-local/issues/1036


firecow:
This should not be merged. It's not how Gitlab handles file permissions.

firecow:
https://github.com/firecow/gitlab-ci-local/pull/1039

This PR will make GCL do exactly what Gitlab Docker Executor does.

I will let you know, when a new version is out, so you can try that out.

hatvik:
> #1039
> 
> This PR will make GCL do exactly what Gitlab Docker Executor does.
> 
> I will let you know, when a new version is out, so you can try that out.
> 
> I'm closing this PR

Yes you may close it, but i dont see how this will fix our issue. 

You are still changing owner:group to 0:0, we do not want to run as root. We run in gitlab CI as the user specified in our docker image, which is a normal user.

Issue #898, opened 2023-06-05
paths is always needed when using cache keyword


aka2dev:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
stages:
  - test
testjob:
  stage: test
  image: alpine:latest
  script:
    - echo "Hello World"
  cache:
    - key: $CI_COMMIT_REF_SLUG
```

**Unexpected behavior**
 ~/projects/github-bugreport  gitlab-ci-local testjob

Using fallback git commit data
Using fallback git remote data
**testjob cache[0].paths must be array**

**Expected behavior**
(By commenting out the cache part in .gitlab-ci.yml)
 ~/projects/github-bugreport  gitlab-ci-local testjob

Using fallback git commit data
Using fallback git remote data
parsing and downloads finished in 21 ms
testjob starting alpine:latest (test)
testjob copied to docker volumes in 489 ms
testjob $ echo "Hello World"
testjob > Hello World
testjob finished in 1.02 s

**Host information**
MacOS 13.3.1 
gitlab-ci-local 4.41.2

**Additional context**
It expects always an array for "paths". In many cases we don't need "paths" in Gitlab. 
"paths" in cache should be optional

firecow:
What does it default to?

firecow:
It copies the files/folders to a hidden folder, and copies it to the build folders that utilize the same cache keys.

firecow:
Ok, `cache: { key: something }` effectively mean, cache disable on gitlab.com

Nothing happens on cache pull or cache push steps in the job log output.

`gitlab-ci-local` should act the same way.

aka2dev:
I agree firecow 👍 if a paths is given, it has to be checked!

aka2dev:
```yaml
stages:
  - test
  
create-cache:
  stage: test
  image: alpine:latest
  script:
    - echo "Hello World" > /some/path/file.txt
  cache:
    - key: $CI_COMMIT_REF_SLUG
        path: /some/path

use-cache:
  stage: test
  image: alpine:latest
  script:
    - cat /some/path/file.txt
    - rm /some/path/file.txt
  cache:
    - key: $CI_COMMIT_REF_SLUG
        policy: pull
  needs:
    - create-cache

use-cache-2:
  stage: test
  image: alpine:latest
  script:
    - cat /some/path/file.txt
  cache:
    - key: $CI_COMMIT_REF_SLUG
        policy: pull
  needs:
    - create-cache
```

[...]

PR #1045, opened 2023-11-22
#898: Add test case

hverlin:


hverlin:
@firecow I think this fix is incorrect, actually. Seems that GitLab CI requires the `paths` to be set:
- without paths, failed: https://gitlab.com/hverlin/gitlab-ci-local-tests/-/pipelines/1081626501
- with paths, worked: https://gitlab.com/hverlin/gitlab-ci-local-tests/-/jobs/5596024562

firecow:
Great, lets close the issue then 👍

Issue #990, opened 2023-09-20
docker-executor jobs shouldn't always run as root:root


firecow:
`USER` set in `Dockerfile` should be used.

Right now `docker create` has `-u 0:0` hardcoded, it shouldn't.

I've verified that gitlab.com runners actually use the linux user specified in `Dockerfile`

![image](https://github.com/firecow/gitlab-ci-local/assets/973602/113e6cee-7de4-43a1-9ac6-ae42e1e60965)

benedikt-bartscher:
Did you try `gitlab-ci-local --umask=false` ?

PR #1048, opened 2023-11-24
Correct docker file permissions and owner using --umask/--no-umask

firecow:
Closes https://github.com/firecow/gitlab-ci-local/issues/990

I've also implemented FF_DISABLE_UMASK_FOR_DOCKER_EXECUTOR, so we can use no-umask on a per job basic.

This is the behaviour i'm trying to reproduce

`--umask` image that has guest user set in Dockerfile
image

`--umask` image that has root user
image

`--no-umask` image that has guest user in Dockerfile
image

`--no-umask` image that has root user
image


firecow:
@hverlin Have a look if you find the time :smile:

hverlin (approved):

Issue #702, opened 2022-12-21
Using of 'DOCKER_HOST' variable will block dind service container to start


iutipikin:
Hi folks. While working with dind jobs I noticed what if you are using well-known docker variables as ci/cd variables e.g. "DOCKER_HOST" to make you dind dokerd working, gitlab-ci-local will fail to start service container (as this assignment will somehow override the one for gitlab-ci-local). Workaround is kind are simple, but in my case I have lots of places to rewrite :(  

Workaround: export DOCKER_HOST=tcp://dind:2375 inside 'script'

**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
job:
  services:
    - name: docker:dind
       alias: dind
  variables:
    DOCKER_HOST: tcp://dind:2375
  script:
    - echo "Heya"
```

**Expected behaviour**
both service and main containers are started, service's dockerd is used in main job container

**Actual behaviour**
`Command failed with exit code 1: docker create -u 0:0 <cmd args> error during connect: Post "http://dind:2375/v1.24/containers/create": dial tcp: lookup dind: no such host`

**Host information**
MacOS 12.6
gitlab-ci-local 4.36.1

**Additional context**
I'm using rancher desktop for docker runtime

firecow:
Please provide a job output of this working on gitlab.com.

firecow:
Closing this for now, feel free to reopen, if you get time to provide us with additional information from a gitlab runner instance.

firecow:
Does this work on remote runners?

trovio-lijn:
I can confirm this as well. I ran a job that had some DOCKER_* vars tonight, and only when I removed (or rather overwrote with `DOCKER_VAR: ""` them all, I saw the dind container starting up in my `watch docker ps` window. 

All of these jobs run fine on gitlab runners (SaaS Premium).

Let me know if I can help out with testing anything. Me trying to fix the code will probably only make things worse 🤪

eljuanchosf:
@firecow - I can confirm that it works on remote runners but not with `gcl`. 

Output is:

```
test        copied to docker volumes in 398 ms
[...]

PR #1062, opened 2023-12-16
Change the way envvars and entrypoints are used by docker containers

firecow:
- Add new test tests/test-cases/dind-no-tls/integration.dind-no-tls.test.ts
- Change the way docker entrypoints are handled
- Change the way envvar's are passed to docker containers

Closes https://github.com/firecow/gitlab-ci-local/issues/702


bcouetil (approved):
You wrote the perfect test, and it is passing, so LGTM

PR #1063, opened 2023-12-19
Upgrade the example to use the new docker compose plugin.

terszt:
According the documentation at https://docs.docker.com/compose/reference/:

> Effective July 2023, Compose V1 stopped receiving updates and is no longer in new Docker Desktop releases. Compose V2 has replaced it and is now integrated into all current Docker Desktop versions.

The example should not use the old docker-compose command (Compose V1).


firecow (approved):
Thanks, looks good. Works on my machine

PR #1066, opened 2023-12-29
argv.ts - respect stateDir when reading dotenv

avoidik:
if stateDir is redefined we'd look up for dotenv file under correct path

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -13,7 +13,7 @@ export class Argv {
             this.map.set(key, value);
         }
 
-        this.injectDotenv(`${this.home}/.gitlab-ci-local/.env`, argv);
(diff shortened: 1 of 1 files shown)


firecow:
I don't believe the .gitlab-ci-local dir located in $HOME has anything todo with --state-dir

avoidik:
You're right, I think this case is only valid when we would set `--home` to some other location than literal `$HOME`. I was trying to match it to a `stateDir` location, but apparently it's a cwd's subdirectory. At this point, if I'd set `--home` to the very same location as in `--cwd`, and change `--stateDir` to `.abc` I will end up with two directories in my project, `.gitlab-ci-local/` and `.abc/`.

PR #1085, opened 2024-02-01
Lowercase `CI_REGISTRY_IMAGE`

PigeonF:
Container images have to be lowercased. A quick read of <https://docs.gitlab.com/ee/ci/variables/predefined_variables.html> does not mention that GitLab automatically lowercases `CI_REGISTRY_IMAGE`, but it does so in practice (see below)

Screenshot 2024-02-01 194921

I don't know if GitLab lowercases only the project path (as in the screenshot), or if it lowercases the `CI_SERVER_HOST` as well. I assume it does both, so I have placed the `toLowerCase` on the outside of the string, instead of after the `this._variables`.


firecow (approved):
Good catch 👍

firecow (approved):
Good catch 👍

PR #1087, opened 2024-02-06
Fix `CI_REGISTRY_IMAGE` (again)

PigeonF:
Hi, while running the master branch locally, I realized that my change in #1085 actually broke `CI_REGISTRY_IMAGE`. The reason is that the full string gets lowercased (i.e. something like `$ci_registry/some/lowercased/path`), which means `$CI_REGISTRY` does not get resolved (because the variable name was lowercased) 😅 . Thats what I get for not testing properly 🤦 .

I have added some tests now to make sure the behaviour is correct now (incidentally, I checked the [gitlab source code](https://gitlab.com/gitlab-org/gitlab/-/blob/0c47378a58c35c234c4eae0d9274763db1cdc7f1/app/models/project.rb#L1291-1295) and it seems they only lowercase the path, not the host, so this code matches the upstream behaviour more closely).

While writing the test I realized that `CI_PROJECT_PATH_SLUG` also needs to be lowercased to match [the upstream behaviour](https://gitlab.com/gitlab-org/gitlab/-/blob/5f43fbc0042ec34e37fbe2531b314b4708a9bbe2/doc/ci/variables/predefined_variables.md#L109).

Sorry for the trouble 😓

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -164,7 +164,7 @@ export class Job {
         predefinedVariables["CI_NODE_INDEX"] = `${opt.nodeIndex}`;
         predefinedVariables["CI_NODE_TOTAL"] = `${opt.nodesTotal}`;
         predefinedVariables["CI_REGISTRY"] = `local-registry.${this.gitData.remote.host}`;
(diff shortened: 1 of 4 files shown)


firecow (approved):
LGTM

PR #1102, opened 2024-02-17
fix: some of before_script/after_script failed to be flattened

### LINKED ISSUE #954: !reference doesn't work in project default before_script/after_script
```yml
---
.hidden:
  script:
    - echo Hello from ${CI_JOB_NAME}

normal_job:
  image: alpine
  before_script:
    - !reference [.hidden, script]  # before_script and after_script under a job work
  script:
    - !reference [.hidden, script]  # This also works
  #after_script:
  # - !reference [.hidden, script]  # Uncommenting the job after_script makes it work

after_script:
  - !reference [.hidden, script]  # before_script and after_script in root of YAML document fail
```
Outputs error:

> normal_job after_script contains non string value

**Expected behavior**
`!reference […]` should work like gitlab CI:

> Executing "step_script" stage of the job script
> Using docker image sha256:7e01a0d0a1dcd9e539f8e9bbd80106d59efbdf97293b3d38f5d7a34501526cdb for alpine with digest alpine@sha256:7144f7bab3d4c2648d7e59409f15ec52a18006a128c733fcff20d3a4a54ba44a ...
> $ echo Hello from ${CI_JOB_NAME}
> Hello from normal_job
> $ echo Hello from ${CI_JOB_NAME}
> Hello from normal_job
> Running after_script
> Running after script...
> $ echo Hello from ${CI_JOB_NAME}
> Hello from normal_job

**Host information**
macOS 12.6.7 Monterey x86_64 intel CPU.
gitlab-ci-local 4.42.0

**Additional context**
Nesting `after_script:` under `default:` does not make a difference:

```yml
#...
default:
  after_script:
    - !reference [.hidden, script]  # before_script and after_script in root of YAML document fail
```

ANGkeith:
make use of `--preview` in test cases to make it easier to write expected test output and also speed up test suites
and fixes some yaml referenced tag's value did not get flattened which closes #954

### DIFF
--- a/src/data-expander.ts
+++ b/src/data-expander.ts
@@ -221,7 +221,8 @@ export function globalVariables (gitlabData: any) {
 
 export function flattenLists (gitlabData: any) {
     traverse(gitlabData, ({parent, key, value}) => {
-        if (!Array.isArray(value) || parent == null || typeof key != "string") return;
-        parent[key] = value.flat(5);
-    });
+        if (parent != null && key != null && Array.isArray(value)) {
+            parent[key] = value.flat(5);
+        }
+    }, {cycleHandling: false});
 }
--- a/tests/test-cases/reference/.gitlab-ci-issue-954.yml
+++ b/tests/test-cases/reference/.gitlab-ci-issue-954.yml
@@ -1,0 +1,14 @@
+---
+.hidden:
+  script:
+    - echo Hello from ${CI_JOB_NAME}
+
+normal_job:
+  image: alpine
+  before_script:
+    - !reference [.hidden, script]  # before_script and after_script under a job work
+  script:
+    - !reference [.hidden, script]  # This also works
+
+after_script:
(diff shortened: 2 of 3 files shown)


src/data-expander.ts
```diff
@@ -220,8 +219,11 @@ export function globalVariables (gitlabData: any) {
 }
 
 export function flattenLists (gitlabData: any) {
-    traverse(gitlabData, ({parent, key, value}) => {
-        if (!Array.isArray(value) || parent == null || typeof key != "string") return;
-        parent[key] = value.flat(5);
-    });
```
ANGkeith, line 226: optional, i found out that disabling the
`cycleHandling` option should fixes the issue as well.

not too sure what you would prefer.

firecow:
But I do have a problem with asserting the entire output is as expected, you end up with assertion roulette.
https://testsmells.org/pages/testsmells.html#AssertionRoulette

firecow (approved):

PR #1120, opened 2024-02-29
fix: cleanup not getting executed

ANGkeith:
```yaml
---
test:
  image: nginx:alpine
  stage: build
  script:
    - sleep 10
```

before 10 seconds is up, send `SIGINT` signal with `<ctrl-c>` 

run `docker ps` and observe that the container is not getting cleaned up

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,11 +15,9 @@ import {Job, cleanupJobResources} from "./job.js";
 
 const jobs: Job[] = [];
 
-process.on("exit", (code) => {
-    cleanupJobResources(jobs).finally(process.exit(code));
-});
-process.on("SIGINT", (_: string, code: number) => {
-    cleanupJobResources(jobs).finally(process.exit(code));
+process.on("SIGINT", async (_: string, code: number) => {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -15,11 +15,9 @@ import {Job, cleanupJobResources} from "./job.js";
 
 const jobs: Job[] = [];
 
-process.on("exit", (code) => {
```
ANGkeith, line 18: removed because `exit` event [Listener functions must only perform synchronous operations.](https://nodejs.org/api/process.html#event-exit)

ANGkeith, line 18: Seemed like the cleanupJobResources() here does nothing, since for event `exit`, [Listener functions must only perform synchronous operations. ](https://nodejs.org/api/process.html#event-exit)

firecow: I think you are right.

But this won't fix that the sleep 10 container isn't cleaned up will it ?

firecow: Nvm, I didn't read the entire change log :smile:

firecow (approved):

PR #1122, opened 2024-03-01
Allow inclusion of the same file multiple times when `inputs:` is used

### LINKED ISSUE #1121: `include:local:inputs` cannot include the same file multiple times
I was testing out the changes from #1115, but it looks like you cannot include the same file multiple times, even though it might be desirable.

Given the following `template.gitlab-ci.yml`

```yml
# template.gitlab-ci.yml
---
spec:
  inputs:
    name:
      type: string
---
"$[[ inputs.name ]]":
  image: busybox:latest
  script:
    - echo "Hello from $[[ inputs.name ]]"
```

The following `.gitlab-ci.yml` works as expected

```yml
# .gitlab-ci.yml
---
include:
  - local: template.gitlab-ci.yml
    inputs:
      name: Foo
```

```console
$ gitlab-ci-local --list
name  description  stage   when        allow_failure  needs
Foo               test    on_success  false  
```

But the following `.gitlab-ci.yml` only shows `Foo`, but there should be both `Foo` and `Bar`.

```yml
# .gitlab-ci.yml
---
include:
  - local: template.gitlab-ci.yml
    inputs:
      name: Foo
  - local: template.gitlab-ci.yml
    inputs:
      name: Bar
```

```console
$ gitlab-ci-local --list
name  description  stage   when        allow_failure  needs
Foo               test    on_success  false  
```

I think the issue is the way `excludedGlobs` is built up when encountering `include` (since this means the file is ignored the second time the include is encountered).

https://github.com/firecow/gitlab-ci-local/blob/0e3602b530bbad48c96006f589022095ae6b58a6/src/parser-includes.ts#L72-L77

PigeonF:
Fixes #1121 .

This is the simplest solution I could think of to fix the issue (though I am not too sure about the reason for the exclusion being used in the first place, so I might have overlooked something).

### DIFF
--- a/src/parser-includes.ts
+++ b/src/parser-includes.ts
@@ -72,7 +72,11 @@ export class ParserIncludes {
                 const files = await globby([value["local"].replace(/^\//, ""), ...excludedGlobs], {dot: true, cwd});
                 for (const localFile of files) {
                     const content = await Parser.loadYaml(`${cwd}/${localFile}`, {inputs: value.inputs || {}});
-                    excludedGlobs.push(`!${localFile}`);
+                    // If the file is included with `inputs:` set, the user might want to include it
+                    // again with different values, so we cannot exclude it.
+                    if(!value["inputs"]) {
+                        excludedGlobs.push(`!${localFile}`);
+                    }
                     includeDatas = includeDatas.concat(await this.init(content, depth, opts));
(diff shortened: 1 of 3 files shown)


PigeonF:
Closing in favor of #1123

PR #1124, opened 2024-03-02
fix: add validation for include local to prevent false valid configurations

ANGkeith:
- `gitlab-ci-local` generates false positive. 
  For the two new test cases for the `include-input` test suites,  
  when running the config via `gitlab-ci-local` it has no issue, but Gitlab is unable to parse these configs.

- smuggled a [one liner](https://github.com/firecow/gitlab-ci-local/pull/1124/commits/03f84733a31ddd50a0db6ce3fa41f661bc9c038f)


firecow (commented):
Looks good to me. Good catch @ANGkeith

firecow (approved):

PR #1132, opened 2024-03-10
feat: when targeting jobs, log padding should only take needs and targeted jobs into account

ANGkeith:
This will ensure that when we're targeting jobs, 

only those jobs that will be executed will be taken into account.

Main motivation of this is because of how we're writing our test-cases,
if we were to add a `very lonooooooooooooooooooooog name` job to a `.gitlab-ci.yml` of a test-suite, it can cause old test
to fail due to the change in the padding


firecow (approved):
Nice addition

PR #1146, opened 2024-03-13
allow gcl to work with duplicated keys but with warnings

### LINKED ISSUE #1047: Find YAML parser that works more like Gitlabs
### Explanation to Developers:

There seems to be a problem with how multiple variable blocks are being parsed in our `.gitlab-ci.yml` configuration file. The issue is causing conflicts for `gitlab-ci-local` while our self-hosted instance of GitLab runs the pipeline just fine.

```yml
---
variables: &security-scan-variables
  YAIR_IMAGE: "registry.example.net/docker-images/yair"

# Additional variables block that may cause parsing issues
variables: &core-variables
  IMAGE_URL: "registry.example.net/example/example/core"
  ARTIFACTORY_IMAGE_URL: "artifactory.example.net/example/core"
  DIRECTORY: "core"

# Jobs definition using the imported variables
security_scan:
  image: *YAIR_IMAGE
  script:
    - echo "Performing security scan on $DIRECTORY"

core_job:
  image: *IMAGE_URL
  script:
    - echo "Running core job with image $ARTIFACTORY_IMAGE_URL"

```

**Expected behavior**
The jobs `security_scan` and `core_job` should execute using their respective variables. As noted earlier, two `variables` declarations at the root level can be problematic in YAML for GitLab CI configuration.

**Host information**
System: Ubuntu 22.04.1
gitlab-ci-local version: 4.45.1

**Containerd binary**
This concern stands regardless of container runtime choice (Docker or Podman), as it pertains to the parsing behavior of the `.gitlab-ci.yml` file within `gitlab-ci-local`.

**Additional context**
With duplicate variables keys at the root level, most parsers would either ignore one set or handle them unpredictably. The appropriate approach is to consolidate all variables under a single variables key or use job-level variables overrides. If gitlab-ci-local differs from GitLab's own CI runner in handling this YAML structure, it may indicate a compatibility issue with gitlab-ci-local.

ANGkeith:
fixes #1047

Just because gitlab supports this

but personally i find that it is very error prone so added a `WARN` message to alert the users

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -259,7 +259,18 @@ export class Parser {
             },
         });
         const schema = yaml.DEFAULT_SCHEMA.extend([referenceType]);
-        const fileData = yaml.loadAll(fileSplitClone.join("\n"), null, {schema}) as any[];
+        let fileData;
+
+        try {
+            fileData = yaml.loadAll(fileSplitClone.join("\n"), null, {schema}) as any[];
+        } catch (e: any) {
+            if (e instanceof yaml.YAMLException && e.reason === "duplicated mapping key") {
+                console.log(chalk`{black.bgYellowBright  WARN } duplicated mapping key detected! Values will be overwritten!`);
+                fileData = yaml.loadAll(fileSplitClone.join("\n"), null, {schema, json: true}) as any[];
+            } else {
+                throw e;
+            }
+        }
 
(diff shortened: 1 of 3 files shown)


firecow (approved):

SansGuidon:
Thank you for the fix even if that makes me mad at Gitlab for tolerating invalid YAML.

PR #1150, opened 2024-03-16
fix: cleanupJobResource

ANGkeith:
`docker rm` and `docker volume rm` supports multiple args. 

this should probably be slightly more efficient.

Not too sure why, but the current cleanup also seemed to leave some `docker volume` undeleted

To replicated the issue:

```yaml
# .gitlab-ci.yml
job0:
  image: bash
  script: echo hi
```

```bash
for i in $(seq 1 9); do
 gitlab-ci-local
done
```

Run `docker volume ls` and observe that there are remaing `xxxxx-tmp` volume that are not cleaned up

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -20,6 +20,9 @@ process.on("SIGINT", async (_: string, code: number) => {
     process.exit(code);
 });
 
+// Graceful shutdown for nodemon
+process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
+
 (() => {
     const packageJson = JSON.parse(fs.readFileSync(path.join(__dirname, "../package.json"), "utf8"));
     yargs(process.argv.slice(2))
--- a/src/index.ts
+++ b/src/index.ts
@@ -41,7 +41,8 @@ process.on("SIGINT", async (_: string, code: number) => {
                     } else {
                         process.stderr.write(chalk`{red ${e.stack ?? e}}\n`);
                     }
-                    cleanupJobResources(jobs).finally(process.exit(1));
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -41,7 +41,8 @@ process.on("SIGINT", async (_: string, code: number) => {
                     } else {
                         process.stderr.write(chalk`{red ${e.stack ?? e}}\n`);
                     }
-                    cleanupJobResources(jobs).finally(process.exit(1));
```
ANGkeith, line 44: with this,

process seemed to exit before cleanup, any idea why ? 


Run this, 
```yaml
cache:
  key:
    files:
      - /foo
  paths:
    - ${CI_PROJECT_DIR}/node_modules
job0:
  image: bash
  script: echo hi
```

and then `docker ps -a`

ANGkeith: Using the following .gitlab-ci.yml
```yaml
cache:
  key:
    files:
      - /foo
  paths:
    - ${CI_PROJECT_DIR}/node_modules
job0:
  image: bash
  script: echo hi
 ```
 
 and running `gitlab-ci-local`
 
 the cleanup works for the `await` syntax but not `finally` syntax

firecow: await is a blocking statement

cleanupJobResources(jobs).finally(process.exit(1)); doesn't block anything

firecow: I imagine the program shuts down, before even getting to the callback function in finally.
So it kindda makes sense.

firecow (approved):

Issue #1171, opened 2024-04-05
Implement --shell-executor-no-image


ANGkeith:
Currently when running the following gitlab-ci.yml, 
```bash
foo:
  script:
    - echo 1
```

it runs as host, and i am hoping to update it to run the default `ruby:3.1`
https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#container-images

 
i would also like to introduce a new cli flag `--host`
to ensure backward compatibility (for eg. [@Interactive](https://github.com/firecow/gitlab-ci-local?tab=readme-ov-file#the-interactive-decorator that only works when running in "host" mode).

I would like to implement this, are you open to this changes ?

PR #1181, opened 2024-04-14
refactor(umask integration tests): use non root images instead of building it

ANGkeith:
a small refactor to the unit test as a pre-requisite to  https://github.com/firecow/gitlab-ci-local/issues/1171

on a side note, should'nt `--umask` be default to `false` instead ? Which should be closer to gitlab-ci default


firecow (approved):

ANGkeith:
@firecow do you have any opinion on this ? 
>  on a side note, should'nt --umask be default to false instead ? Which should be closer to gitlab-ci default

PR #1186, opened 2024-04-15
feat: implement shell executor no image

### LINKED ISSUE #1171: Implement --shell-executor-no-image

ANGkeith:
closes #1171

merge https://github.com/firecow/gitlab-ci-local/pull/1186 in first

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -57,6 +57,10 @@ export class Argv {
 
         this.injectDotenv(`${this.home}/.gitlab-ci-local/.env`, argv);
         this.injectDotenv(`${this.cwd}/.gitlab-ci-local-env`, argv);
+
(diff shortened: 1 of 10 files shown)


firecow (approved):
I just :heart: the time you take in writing the tests :smile: 

It makes it so easy to approve

Issue #1178, opened 2024-04-08
Gcl use my username as host in includes


comm1x:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
include:
  - project: 'myrepo/myapp'
    ref: 'master'
    file:
      - 'config.yml'
```

**Expected behavior**
GCL download config.yml from Gitlab and use it.

**Actual behavior**
GCL use my username as `remote.host` and failed with error

```
Project include could not be fetched 
fatal: unable to access 'https://p.shorokhov/myrepo/myapp.git/': Could not resolve host: p.shorokhov
```

**Host information**
Macos
gitlab-ci-local 4.47.0

**Containerd binary**
Docker

Gcl failed in this `git clone` command: https://github.com/firecow/gitlab-ci-local/blob/c6d78471ca2b4305572cc31b131f96cb6416bd5e/src/parser-includes.ts#L190

firecow:
Paste your `~/.gitlab-ci-local/variables.yml`, `env | grep GCL_` and `<project>/.gitlab-ci-local-env` (Mask your secrets)

comm1x:
`~/.gitlab-ci-local` not exists

`env | grep GCL_` empty

$project/.gitlab-ci-local-env: `VARIABLE="CI_JOB_TOKEN=... CI_PIPELINE_SOURCE=merge_request_event CI_MERGE_REQUEST_ID=1"`

git remote -v:
```
origin  ssh://git@gitlab.mydomain.ru:5555/myrepo/myapp.git (fetch)
origin  ssh://git@gitlab.mydomain.ru:5555/myrepo/myapp.git (push)
```

firecow:
You must have some setting somewhere in the `<project>/.git`, global git settings, or gitlab-ci-local variables or env settings, that makes `gitlab-ci-local` resolves https://p.shorokhov as being the place to go.

comm1x:
I have `User p.shorokhov` in `.ssh/config` in gitlab host. Can't find any other entries.

ANGkeith:
@comm1x 

Are you showing your full `git remote -v` ? 

Is this the error you're seeing? 
![image](https://github.com/firecow/gitlab-ci-local/assets/28011783/faaba21b-1972-49d8-a565-a2c61eedd9b7)

ANGkeith:
https://github.com/firecow/gitlab-ci-local/blob/master/src/git-data.ts#L85

@firecow 
[...]

Issue #921, opened 2023-07-06
Better heuristic for resolving project urls


nolange:
**Is your feature request related to a problem? Please describe.**

My CI File includes a few project references, well... those ideally would be resolved.
The git repository has 2 remote servers, one being the gitlab server, one a "dump" backup,
gitlab-ci-local tries to fetch the referenced project from the non-gitlab server - where it does not exist.

**Describe the solution you'd like**
When the active branch is setup to track a remote branch,
then gitlab-ci-local should  prefer **this** URL instead of whatever it does now.
This is my setup (branch tracks upstream gitlab server branch), so gitlab-ci-local would then "just work".

(For weirder setups some commandline arguments might be necessary).

**Describe alternatives you've considered**
Removing all but the the correct server as a workaround.

firecow:
How would gitlab-ci-local be able to chose between your two remotes exactly ?

nolange:
> How should gitlab-ci-local determine which of your two remotes to use exactly ?

Look at the current branch, see it it has an upstream branch (default push/pull target). If so, then use the remote from the upstream branch.

firecow:
Which git commands output should be parsed?
Paste an example out here.

nolange:
If you are in a git repository, you get the upstream branch (or an error message) with:

``` bash
git rev-parse --abbrev-ref @{upstream} 
```

would return origin/main in my example

you can then get the url with

``` bash
git remote get-url origin
```

you can of course get the information also from the files directly:

```
# points to the current branch
.git/HEAD: "ref: refs/heads/main"
# remote and branch configuration
.git/config
```

Example .git/config

``` ini
[remote "test"]
	url = git@10.99.0.1:/backup/project
	fetch = +refs/heads/*:refs/remotes/test/*
[remote "origin"]
	url = git@gitlab.mystuff.com:software/project.git
	fetch = +refs/heads/*:refs/remotes/origin/*
[branch "main"]
	remote = origin
	merge = refs/heads/main
```

martin-142:
I have the same issue using a git repo with two different remotes, gitlab-ci-local tries to fetch the include files from `git@gitlab.com` url, instead of using the custom GitLab instance URL which is the `origin`. This because gitlab-ci-local uses the first remote url from the command `git remote -v`.

The command that @nolange mention works if the user already set the tracking information with `--set-upstream-to`, this is not my case, so I get an error. I don't know the behavior if the current branch exists in both remotes repos...

[...]

Issue #1083, opened 2024-02-01
gitlab-ci-local includes remote ssh port


WladyX:
**Is your feature request related to a problem? Please describe.**
gitlab-ci-local does not work as expected ( by me :D ) with custom ssh ports, see #1081
if gitlab ssh is using a custom port (defined in ~/.ssh/config) gitlab-ci-local is not able to pull the includes.

**Describe the solution you'd like**
ideally we would be able to specify the ssh remote port with a flag like : `--remote-ssh-port`

Thank you for sharing this project!

WladyX:
This looks resolved now, thank you @ANGkeith and @firecow !

PR #1202, opened 2024-05-02
fix: parsing of basic auth remote-url

ANGkeith:
not too sure how to fix the code smell though

possible fix for #1178 and #921

### DIFF
--- a/src/git-data.ts
+++ b/src/git-data.ts
@@ -89,16 +89,17 @@ export class GitData {
                 gitRemoteMatch = /(?<schema>https?):\/\/(?:(\w+):([\w-]+)@)?(?<host>[^/:]+):?(?<port>\d+)?\/(?<group>\S+)\/(?<project>\S+)\.git/.exec(gitRemote); // regexr.com/7ve8l
                 assert(gitRemoteMatch?.groups != null, "git remote get-url gcl-origin 2> /dev/null || git remote get-url origin");
 
+                let port = "443";
                 if (gitRemoteMatch.groups.schema === "https") {
-                    this.remote.port = gitRemoteMatch.groups.port ?? "443";
+                    port = gitRemoteMatch.groups.port ?? "443";
                 } else if (gitRemoteMatch.groups.schema === "http") {
-                    this.remote.port = gitRemoteMatch.groups.port ?? "80";
+                    port = gitRemoteMatch.groups.port ?? "80";
                 }
                 this.remote.host = gitRemoteMatch.groups.host;
                 this.remote.group = gitRemoteMatch.groups.group;
                 this.remote.project = gitRemoteMatch.groups.project;
                 this.remote.schema = gitRemoteMatch.groups.schema;
-
+                this.remote.port = port;
             } else if (gitRemote.startsWith("ssh://")) {
                 gitRemoteMatch = /(?<schema>ssh):\/\/(\w+)@(?<host>[^/:]+):?(?<port>\d+)?\/(?<group>\S+)\/(?<project>\S+)\.git/.exec(gitRemote); // regexr.com/7vjq4
                 assert(gitRemoteMatch?.groups != null, "git remote get-url gcl-origin 2> /dev/null || git remote get-url origin");
--- a/src/git-data.ts
+++ b/src/git-data.ts
@@ -83,11 +83,19 @@ export class GitData {
     private async initRemoteData (cwd: string, writeStreams: WriteStreams): Promise<void> {
         try {
             let gitRemoteMatch;
-            const {stdout: gitRemote} = await Utils.spawn(["bash", "-c", "git remote get-url gcl-origin 2> /dev/null || git remote get-url origin"], cwd);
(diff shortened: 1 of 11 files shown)


firecow:
https://sonarcloud.io/dashboard?id=firecow_gitlab-ci-local&pullRequest=1202 Have a look at the code smells @ANGkeith 

I'll mark the security findings as safe.

ANGkeith:
> Have a look at the code smells @ANGkeith

yeah, i am saw those code smells, not too sure how to fix it : x 


some of the named groups are used, but the sonarqube seemed to fail to detect it.

firecow:
> > Have a look at the code smells @ANGkeith
> 
> yeah, i am aware of those code smells, not too sure how to fix it : x
> 
> They (named groups) are used, but the sonarqube seemed to fail to detect it.

Oh... Then we shall ignore. I didn't think SonarQube could produce false positives. Never experienced that.

firecow:
@ANGkeith I've fixed the code smells with https://github.com/firecow/gitlab-ci-local/pull/1202/commits/6f89d910437cc27036d5e8a1c88ef637bf6a1d95

src/git-data.ts
```diff
@@ -82,24 +82,44 @@ export class GitData {
...
         try {
-            const {stdout: gitRemote} = await Utils.spawn(["git", "remote", "-v"], cwd);
-            const gitRemoteMatch = /(?<schema>git|https?)(?::\/\/|@)(?<host>[^:/]*)(:(?<port>\d+)\/|:|\/)(?<group>.*)\/(?<project>[^ .]+)(?:\.git)?.*/.exec(gitRemote);
-
-            assert(gitRemoteMatch?.groups != null, "git remote -v didn't provide valid matches");
-
-            this.remote.schema = gitRemoteMatch.groups.schema;
-            if (this.remote.schema === "git") {
+            let gitRemoteMatch;
+            const {stdout: gitRemote} = await Utils.spawn(["bash", "-c", "git remote get-url gcl-origin 2> /dev/null || git remote get-url origin"], cwd);
```
firecow, line 86: Can you explain the logic behind this?

ANGkeith: > Can you explain the logic behind this?
https://github.com/firecow/gitlab-ci-local/issues/921#issuecomment-2090224041
`gcl-origin` is to for people to customize their own "remote" in the event that `origin` is not what they want

ANGkeith: > If bash is really needed Utils.bash should be used.
> Keep in mind also, that bash is a lot slower, than just calling plain ol' shell cmd's

Ok, got it, wasnt aware of this.

ANGkeith: > Wouldn't this be cleaner?
LGTM

ANGkeith:
> @ANGkeith I've fixed the code smells with [6f89d91](https://github.com/firecow/gitlab-ci-local/commit/6f89d910437cc27036d5e8a1c88ef637bf6a1d95)

thanks!

src/git-data.ts
```diff
@@ -82,24 +82,44 @@ export class GitData {
...
+                this.remote.host = gitRemoteMatch.groups.host;
+                this.remote.group = gitRemoteMatch.groups.group;
+                this.remote.project = gitRemoteMatch.groups.project;
+                this.remote.schema = gitRemoteMatch.groups.schema;
                 this.remote.port = gitRemoteMatch.groups.port ?? "22";
+            } else {
+                gitRemoteMatch = /(?<username>\S+)@(?<host>[^:]+):(?<group>\S+)\/(?<project>[^ .]+)\.git/.exec(gitRemote); // regexr.com/7vjoq
+                assert(gitRemoteMatch?.groups != null, "git remote get-url gcl-origin 2> /dev/null || git remote get-url origin");
+
+                const {stdout: port} = await Utils.bash(`ssh -G ${gitRemoteMatch.groups.username}@${gitRemoteMatch.groups.host} | grep '^port' | awk '{print $2}'`);
```
ANGkeith, line 116: btw, for the fix for #1083

This should set the correct port number to use for 
https://github.com/firecow/gitlab-ci-local/blob/250bab2232a65f89f6813ebc67d99fe2b5b00dbf/src/parser-includes.ts#L203

src/git-data.ts
ANGkeith: > Can you explain the logic behind this?
> 
> ```ts
> let gitRemote;
> try {
>     const res = await Utils.spawn(["git", "remote", "get-url", "gcl-origin"], cwd);
>     gitRemote = res.stdout;
> } catch {
>     const res = await Utils.spawn(["git", "remote", "get-url", "origin"], cwd);
>     gitRemote = res.stdout;
> }
> ```
> 
> Wouldn't this be cleaner? (I don't like bash magic) 😄

yeah lgtm,though i am a bash magician 😂

firecow: Haha, yeah... But it ramps up, spawning that many bash instances :smile:

firecow (approved):

PR #1203, opened 2024-05-05
Implement pull policy

firecow:


src/job.ts
```diff
@@ -829,14 +829,22 @@ export class Job {
...
+        const actualPull = async () => {
             const time = process.hrtime();
             await Utils.spawn([this.argv.containerExecutable, "pull", imageToPull]);
             const endTime = process.hrtime(time);
             writeStreams.stdout(chalk`${this.formattedJobName} {magentaBright pulled} ${imageToPull} in {magenta ${prettyHrtime(endTime)}}\n`);
             this.refreshLongRunningSilentTimeout(writeStreams);
+        };
+
+        if (pullPolicy === "always") {
+            return actualPull();
```
ANGkeith, line 842:
```suggestion
          await actualPull();
          return 
```

firecow: Whats the different?

ANGkeith: My initial motivation was to be more consistent, 

since for the flow when `pullPolicy !== "always:`, 

```javascript
        try {
            await Utils.spawn([this.argv.containerExecutable, "image", "inspect", imageToPull]);
        } catch (e: any) {
            await actualPull();
        }
 ```

firecow: @ANGkeith I've applied it.

ANGkeith: was doing some readings as well, seemed like there's some subtle difference with with the stacktrace. 

`return promise` vs `return await promise`

https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/errorhandling/returningpromises.md#code-example-anti-pattern-calling-async-function-without-awaiting

ANGkeith (approved):

PR #1204, opened 2024-05-05
Use Utils.spawn over Utils.bash

firecow:


ANGkeith:
i am hoping to update the Utils.spawn
from
```javascript
    static spawn (cmdArgs: string[], cwd = process.cwd()): Promise<{stdout: string; stderr: string}> {
        return execa(cmdArgs[0], cmdArgs.slice(1), {cwd});
    }
```

```javascript
async function sh (script)  {
    return execa("sh", ["-c", script]);
}
```

what do you think ? 

main motivation is because, the current `spawn` does not allow pipelines, making it difficult to convert bash to spawn for commands like 
https://github.com/firecow/gitlab-ci-local/blob/250bab2232a65f89f6813ebc67d99fe2b5b00dbf/src/parser-includes.ts#L203

ANGkeith (approved):

PR #1218, opened 2024-05-14
fix: git remote get-url origin for git:// can contain dots

firecow:
The regexr comment actually had the solution. @ANGkeith 
// regexr.com/7vjoq

Our current solution dosn't work if projects containers a `.`

`git@github.com:fire/cow/example.com.git`


SlyngDK (approved):

ANGkeith (approved):

sgaist:
Nice !
I was about to submit a patch for it :-)

Issue #1217, opened 2024-05-13
Add support for predefined runner instance class


nicDamours:
**Is your feature request related to a problem? Please describe.**
I was trying to debug some performance issues with my pipeline and I could not set the cpu and memory limit of the docker runner that gitlab-ci-local is using. I was able to make it work by editing the source and specifying those values manually, but it was quite the challenge.

**Describe the solution you'd like**
Could it be possible to add an option to emulate the resources of a gitlab shared runner ? For example, we could use the option `--emulate-runner=saas-linux-small-amd64` which could set the memory and CPUs limit to those of the `saas-linux-small-amd64` instance. This would help debug performance issues.

The list of the available gitlab runner hosted by gitlab can be found here. 
https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html

**Describe alternatives you've considered**
I looked into the `ulimit` option, but it seams quite limited compared to `--memory=`, `--cpus=`, etc. 

As stated previously, I was able to define those values manually and emulate gitlab's small runner. 

**Additional context**
My crude solution was to add the following line to the `execScript` function of the `job.ts` file. 

```javascript
dockerCmd += "--memory=8096m --kernel-memory=8096m --cpus=2 ";
```

I will try to implements the feature and submit a pull-request.

firecow:
It should be possible.

PR #1219, opened 2024-05-14
[FEATURE] #1217 Added feature to emulate gitlab hosted runner using the `--container-emulate` option with the name of one of the gitlab preset.

nicDamours:
### summary
Added an option to prefill `memory`, `kernel-memory` and  `cpus` docker options based on the configuration of Gitlab's hosted runner. This is useful when attempting to debug performance issue. By default, no limit is applied, the host resources are all available. 

When passing the option `--container-emulate=` with the name of a gitlab hosted runner, the docker resources is limited to the resource available for this class of hosted runner. As of this PR, those are the values

| name | memory | CPU | 
| --- | --- | --- |
| `saas-linux-small` | 8 GB | 2 |
| `saas-linux-medium` | 16 GB | 4 |
| `saas-linux-large` | 32 GB | 8 |
| `saas-linux-xlarge` | 64 GB | 16 |
| `saas-linux-2xlarge` | 128 GB | 32 |

[see the gitlab documentation here](https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#machine-types-available-for-linux---x86-64)

Of course, you are also limited to the host resources. You cannot have 128 GB of RAM on a host that only has 16 GB.

As of right now, I've only added support for the base linux machine. This does not include runner with a GPUs. This could be a nice amelioration in the futur. The runners names does not contains the architecture, as this is based on the host CPU, and cannot be emulated.

### Usage

You can pass an additional option `--containter-emulate` with one of the previously mentionned runner name. For example, using the name `saas-linux-small` will limit the container resource to 8 GB or RAM and 2 CPUs. 

```
gitlab-ci-local --cwd /path/to/.gitlab.yml --container-emulate=saas-linux-small
```

You can check the memory limit and CPU usage using the `docker stats {container_id}` command.

image

This PR Closes https://github.com/firecow/gitlab-ci-local/issues/1217

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -240,6 +240,10 @@ export class Argv {
         return this.map.get("containerMacAddress") ?? null;
     }
 
+    get emulate (): string | null {
+        return this.map.get("emulate") ?? null;
+    }
+
     get concurrency (): number | null {
         const concurrency = this.map.get("concurrency");
         if (!concurrency) return null;
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -240,8 +240,8 @@ export class Argv {
         return this.map.get("containerMacAddress") ?? null;
     }
 
-    get emulate (): string | null {
-        return this.map.get("emulate") ?? null;
+    get containerEmulate (): string | null {
+        return this.map.get("container-emulate") ?? null;
     }
 
     get concurrency (): number | null {
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -241,7 +241,7 @@ export class Argv {
     }
 
     get containerEmulate (): string | null {
-        return this.map.get("container-emulate") ?? null;
+        return this.map.get("containerEmulate") ?? null;
     }
 
     get concurrency (): number | null {
--- a/src/gitlab-preset.ts
+++ b/src/gitlab-preset.ts
@@ -1,9 +1,9 @@
 export enum GitlabRunnerPreset {
-    AMD_64_SMALL = "saas-linux-small-amd64",
-    AMD_64_MEDIUM = "saas-linux-medium-amd64",
-    AMD_64_LARGE = "saas-linux-large-amd64",
-    AMD_64_XLARGE = "saas-linux-xlarge-amd64",
-    AMD_64_2XLARGE = "saas-linux-2xlarge-amd64",
+    AMD_64_SMALL = "saas-linux-small",
+    AMD_64_MEDIUM = "saas-linux-medium",
+    AMD_64_LARGE = "saas-linux-large",
+    AMD_64_XLARGE = "saas-linux-xlarge",
+    AMD_64_2XLARGE = "saas-linux-2xlarge",
 }
 
 export const GitlabRunnerPresetValues: string[] = Object.values(GitlabRunnerPreset);
--- a/src/gitlab-preset.ts
+++ b/src/gitlab-preset.ts
@@ -1,0 +1,25 @@
+export enum GitlabRunnerPreset {
+    AMD_64_SMALL = "saas-linux-small-amd64",
+    AMD_64_MEDIUM = "saas-linux-medium-amd64",
+    AMD_64_LARGE = "saas-linux-large-amd64",
+    AMD_64_XLARGE = "saas-linux-xlarge-amd64",
+    AMD_64_2XLARGE = "saas-linux-2xlarge-amd64",
+}
+
+export const GitlabRunnerPresetValues: string[] = Object.values(GitlabRunnerPreset);
+
+export const GitlabRunnerMemoryPresetValue: Record<string, number> = {
+    [GitlabRunnerPreset.AMD_64_SMALL]: 8096,
+    [GitlabRunnerPreset.AMD_64_MEDIUM]: 16384,
+    [GitlabRunnerPreset.AMD_64_LARGE]: 32768,
+    [GitlabRunnerPreset.AMD_64_XLARGE]: 65536,
+    [GitlabRunnerPreset.AMD_64_2XLARGE]: 131072,
+};
+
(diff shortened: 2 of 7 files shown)


src/gitlab-preset.ts
```diff
@@ -0,0 +1,25 @@
...
+    [GitlabRunnerPreset.AMD_64_2XLARGE]: 131072,
+};
+
+export const GitlabRunnerCPUsPresetValue: Record<string, number> = {
+    [GitlabRunnerPreset.AMD_64_SMALL]: 2,
+    [GitlabRunnerPreset.AMD_64_MEDIUM]: 4,
+    [GitlabRunnerPreset.AMD_64_LARGE]: 8,
+    [GitlabRunnerPreset.AMD_64_XLARGE]: 16,
+    [GitlabRunnerPreset.AMD_64_2XLARGE]: 32,
+};
```
ANGkeith, line 25:
```suggestion
};

```

ANGkeith: nit: misssing new line

tests/test-cases/gitlab-runner-emulation/integration.emulate-runner.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+        expect(true).toBe(false);
+    } catch (e: unknown) {
+        expect((e as Error).message).toContain("Invalid gitlab runner to emulate.");
+    }
+});
+
+test("should contains memory config when emulating valid runner", async () => {
+    // given a spy of the bash utils
+    const bashSpy = jest.spyOn(Utils, "bash").mockResolvedValue({stdout: "abcde12345", stderr: "", exitCode: 0});
```
ANGkeith, line 35: Lets use the helper for mocking
https://github.com/firecow/gitlab-ci-local/blob/70f4b74ee74edeb3c6806a926169ed41b334f6fd/tests/test-cases/remote-variables-file/integration.project-variables-file.test.ts#L13

src/gitlab-preset.ts
```diff
@@ -0,0 +1,25 @@
...
+    AMD_64_MEDIUM = "saas-linux-medium-amd64",
+    AMD_64_LARGE = "saas-linux-large-amd64",
+    AMD_64_XLARGE = "saas-linux-xlarge-amd64",
+    AMD_64_2XLARGE = "saas-linux-2xlarge-amd64",
+}
+
+export const GitlabRunnerPresetValues: string[] = Object.values(GitlabRunnerPreset);
+
+export const GitlabRunnerMemoryPresetValue: Record<string, number> = {
+    [GitlabRunnerPreset.AMD_64_SMALL]: 8096,
```
ANGkeith, line 12: nit: personally i would prefer `8 * 1024` to save on the arithmetic

nicDamours: I agree, it helps alot with the readability of the constants.

src/index.ts
```diff
@@ -257,6 +258,11 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
             description: "Container MAC address (e.g., aa:bb:cc:dd:ee:ff)",
             requiresArg: false,
         })
+        .option("emulate", {
```
ANGkeith, line 261: maybe one last change, 

`--emulate` -> `--container-emulate` 

Thinking about having the convention of `container` prefix for all container related options

nicDamours: I like the idea of a convention, i will update the code accordingly.

src/index.ts
```diff
@@ -257,6 +258,11 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
...
+            type: "string",
+            description: "The name of a gitlab hosted runner to emulate",
```
ANGkeith, line 263: https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#machine-types-available-for-linux---x86-64

let's add the link in the description too,  for a slightly better ux.

ANGkeith:
thanks for the contribution !

src/argv.ts
```diff
@@ -240,6 +240,10 @@ export class Argv {
         return this.map.get("containerMacAddress") ?? null;
     }
 
+    get containerEmulate (): string | null {
+        return this.map.get("container-emulate") ?? null;
```
ANGkeith, line 244:
```suggestion
        return this.map.get("containerEmulate") ?? null;
```

ANGkeith (approved):

firecow:
@nicDamours I don't see the purpose of the amd64 postfix.

nicDamours:
> @nicDamours I don't see the purpose of the amd64 postfix.

@firecow it's the name given by gitlab for those specific runner: https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#machine-types-available-for-linux---x86-64

It's not essential for this feature but it helps linking gitlab's runner to the preset defined here.

firecow:
> > @nicDamours I don't see the purpose of the amd64 postfix.
> 
> @firecow it's the name given by gitlab for those specific runner: https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#machine-types-available-for-linux---x86-64
> 
> It's not essential for this feature but it helps linking gitlab's runner to the preset defined here.

I can imagine people with ARM processors will be very confused, when they realise that using such a preset, actually works :smile:

nicDamours:
I can see how that would be confusing. I will remove the amd64 prefix :slightly_smiling_face:

src/gitlab-preset.ts
```diff
@@ -0,0 +1,25 @@
+export enum GitlabRunnerPreset {
+    AMD_64_SMALL = "saas-linux-small",
```
ANGkeith, line 2: lets remove the `amd` here too

ANGkeith (approved):

PR #1221, opened 2024-05-15
feat: implement rules:changes and rules:changes:path

### LINKED ISSUE #395: Support rules:changes
In Gitlab CI you can add the keyword "rules:changes" for a job that should only be executed if specified file changed.: https://docs.gitlab.com/ee/ci/yaml/#ruleschanges

Example:
```
build:
  script:
    - echo "building"
  rules:
    - if: $CI_COMMIT_BRANCH == 'master'
      changes:
        - src/**/*
```

What i want to achieve with that: Assuming you multiple services and all of them will be compiled in a big pipeline. So if i only change something in one of the services i have to wait since the whole pipeline finished building. With "rules:changes" i want to achieve something similar other tools like skaffold do: Only build that service / container where changes in code / configuration happend.

Actual problem: gitlab-ci-local seems to completely ignore that setting. I dont get an error but the job runs regardless of whether changes take place or not.

ANGkeith:
closes #395

#### added new dependencies
npm i micromatch
npm i --save-dev @types/micromatch

### DIFF
--- a/src/git-data.ts
+++ b/src/git-data.ts
@@ -79,6 +79,10 @@ export class GitData {
         }
     }
 
+    static changedFiles (defaultBranch: string) {
+        return Utils.syncSpawn(["git", "diff", "--name-only", defaultBranch]).stdout.split("\n");
+    }
+
     private async initRemoteData (cwd: string, writeStreams: WriteStreams): Promise<void> {
         try {
             let gitRemoteMatch;
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -6,7 +6,9 @@ import base64url from "base64url";
 import execa from "execa";
 import assert from "assert";
 import {CICDVariable} from "./variables-from-files";
+import {GitData} from "./git-data";
 import globby from "globby";
+import micromatch from "micromatch";
 import axios from "axios";
 
 type RuleResultOpt = {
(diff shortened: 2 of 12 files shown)


src/utils.ts
```diff
@@ -35,6 +38,10 @@ export class Utils {
         return execa(cmdArgs[0], cmdArgs.slice(1), {cwd});
     }
 
+    static syncSpawn (cmdArgs: string[], cwd = process.cwd()): {stdout: string; stderr: string} {
```
ANGkeith, line 41: created a sync version of `Utils.spawn`, because `Utils.getRulesResult` is not async

src/git-data.ts
```diff
@@ -79,6 +79,11 @@ export class GitData {
         }
     }
 
+    static changedFiles (defaultBranch: string) {
+        return Utils.syncSpawn(["sh", "-c", `git diff --name-only ${defaultBranch}`])
+            .stdout.split("\n");
+    }
+
```
firecow, line 86:
```suggestion
    static changedFiles (defaultBranch: string) {
        return Utils.syncSpawn(["git", "diff", "--name-only", defaultBranch]).stdout.split("\n");
    }

```

Issue #1225, opened 2024-05-17
Unable to run with --network host


octane100:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
image: alpine

job:
  script:
    - ip r
```
With gitlab-ci-local 4.49.0, the following error is reported when running with parameter `--network host` : 
```
$gitlab-ci-local --network=host
job starting alpine:latest (test)
job copied to docker volumes in 590 ms
Error: Command failed with exit code 1: docker network connect host c85b4b31fd69350dd7538d4653db289bd5bb2b3d2c4302ab096710fbdf1593f7
Error response from daemon: container cannot be disconnected from host network or connected to host network
    at makeError (/snapshot/firecow-gitlab-ci-local/node_modules/execa/lib/error.js:60:11)
    at handlePromise (/snapshot/firecow-gitlab-ci-local/node_modules/execa/index.js:118:26)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Job.execScripts (/snapshot/firecow-gitlab-ci-local/src/job.ts:737:17)
    at Job.execPreScripts (/snapshot/firecow-gitlab-ci-local/src/job.ts:602:36)
    at Job.start (/snapshot/firecow-gitlab-ci-local/src/job.ts:499:9)
    at /snapshot/firecow-gitlab-ci-local/node_modules/p-map/index.js:57:22
```

**Expected behavior**
Being able to run with host networking

**Host information**
Ubuntu 22.04
gitlab-ci-local 4.49.0

**Containerd binary**
docker.io 24.0.5-0ubuntu1~22.04.1

**Additional context**
The same command is working as expected with gitlab-ci-local 4.48.2.
Regression may have been introduced by #1205

PigeonF:
I can reproduce your issue, and you are right about the PR that introduced the issue.

Reading through <https://docs.docker.com/network/>, the same error happens for

```console
$ gitlab-ci-local --network=host
$ gitlab-ci-local --network=none
```

Additionally, the following syntax works to connect to a container directly using docker CLI

```console
$ docker run -d --name redis redis --bind 127.0.0.1
$ docker run --rm -it --network container:redis redis redis-cli -h 127.0.0.1 PING
```

but when used with `docker connect`, it seems the same syntax does not work

```console
$ docker run -d --name redis redis --bind 127.0.0.1
$ gitlab-ci-local --network container:redis
parsing and downloads finished in 48 ms
[...]

PR #1226, opened 2024-05-17
Handle `host` and `none` networks

PigeonF:
Related to #1225

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -693,6 +693,14 @@ export class Job {
                 dockerCmd += `--cpus=${cpuConfig} `;
             }
 
+            // host and none networks have to be specified using --network,
+            // since they cannot be used with `docker network connect`.
+            for (const network of this.argv.network) {
+                if (["host", "none"].includes(network)) {
+                    dockerCmd += `--network ${network} `;
+                }
+            }
+
             dockerCmd += `--volume ${buildVolumeName}:/gcl-builds `;
             dockerCmd += `--volume ${tmpVolumeName}:${this.fileVariablesDir} `;
             dockerCmd += "--workdir /gcl-builds ";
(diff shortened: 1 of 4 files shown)


PigeonF:
I am not sure how to test connecting to custom networks, since it would require setting up the docker networks in the test. Is there some setup/teardown mechanics for this?

nicDamours:
@PigeonF You don't have to actually validate that the docker container connects to the proper network. It is out of the scope of this library to test docker itself. However, you can validate that you correctly pass the needed argument to the docker command. 

To do so, you can use the return value of the  `initBashSpy`, and use the `expect(bashSpy).toHaveBeenLastCalledWith` function to assert the content of the command. 

for example: 
```typescript
expect(bashSpy).toHaveBeenLastCalledWith(expect.stringContaining("--network someNetwork"), expect.any(String));
```
[Here is an example of this being done in another test](https://github.com/firecow/gitlab-ci-local/blob/b7851f56963c0692270a11978da16462cf9aa42d/tests/test-cases/gitlab-runner-emulation/integration.emulate-runner.test.ts#L46)

You can also test the `Utils.spawn` command you use. In order to do that, you would need to edit the `initSpawnSpy` to also return the spy and then use the `toHaveBeenLastCalledWith` function, again, to validate the parameters. For example: 

```typescript
expect(spawnSpy).toHaveBeenCalledWith(expect.any(String), "network", "connect", "theNetwork", expect.any(String))
```

PigeonF:
Thank you for the help! I have added some tests and have discovered, that my solution is not in fact a solution 🤦 . I have pushed the tests for now, while I am working on a fix.

PigeonF:
I think this should cover the relevant cases now - though I don't know how well behaved docker is in some of these combinations (i.e. why can we not use `--network host --network custom` when calling `docker container create`, but are allowed to do it in two steps with `docker network create --network host; docker network connect custom`?)

firecow:
@PigeonF Is this one ready for review?
I'm thinking about releasing a new version in the near future.

firecow (approved):

PR #1167, opened 2024-03-31
feat: add json schema validation with ajv

ANGkeith:

PR #1229, opened 2024-05-22
Implement plain `parallel:` keyword

### LINKED ISSUE #1227: $CI_NODE_INDEX defaults to the string "null" when not in a parallel job, expected the empty string
```yml
---
example:
  script:
    - echo $CI_NODE_INDEX
    - echo $CI_NODE_TOTAL
    - echo $This_variable_does_not_exist
```

**Expected behavior**
Snippet of gitlab.com job output:
```
$ echo $CI_NODE_INDEX
$ echo $CI_NODE_TOTAL
1
$ echo $This_variable_does_not_exist
Cleaning up project directory and file based variables
00:01
Job succeeded
```

**Host information**
Ubuntu 20.04
gitlab-ci-local 4.49.0

**Containerd binary**
Are you using docker or podman, or any other relevant containerization tool. - No

**Additional context**
Looks like the specific lines that set the variable are here:

https://github.com/firecow/gitlab-ci-local/blob/d6378db7dc6975c92356d16be0d7b9f563f7d9d9/src/job.ts#L29

https://github.com/firecow/gitlab-ci-local/blob/d6378db7dc6975c92356d16be0d7b9f563f7d9d9/src/job.ts#L172

PigeonF:
See <https://docs.gitlab.com/ee/ci/yaml/#parallel>.

When writing the test I saw that there was some verification of the schema (e.g. I could not set `parallel: 1`), but when running `gitlab-ci-local` directly it works fine. Is this intended behaviour? There is now an edge case here, which I cannot test for at the moment: Setting `parallel: 1` works on <https://gitlab.com>, but I cannot write a test for it because it does not match the official schema (which is apparently wrong, I have filed <https://gitlab.com/gitlab-org/gitlab/-/issues/462979>).

This fixes #1227 .

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -169,7 +169,9 @@ export class Job {
         predefinedVariables["CI_ENVIRONMENT_NAME"] = this.environment?.name ?? "";
         predefinedVariables["CI_ENVIRONMENT_SLUG"] = this.environment?.name?.replace(/[^a-z\d]+/ig, "-").replace(/^-/, "").slice(0, 23).replace(/-$/, "").toLowerCase() ?? "";
         predefinedVariables["CI_ENVIRONMENT_URL"] = this.environment?.url ?? "";
-        predefinedVariables["CI_NODE_INDEX"] = `${opt.nodeIndex}`;
+        if (opt.nodeIndex !== null) {
+            predefinedVariables["CI_NODE_INDEX"] = `${opt.nodeIndex}`;
+        }
         predefinedVariables["CI_NODE_TOTAL"] = `${opt.nodesTotal}`;
         predefinedVariables["CI_REGISTRY"] = `local-registry.${this.gitData.remote.host}`;
         predefinedVariables["CI_REGISTRY_IMAGE"] = `$CI_REGISTRY/${this._variables["CI_PROJECT_PATH"].toLowerCase()}`;
--- a/src/parallel.ts
+++ b/src/parallel.ts
@@ -1,12 +1,16 @@
 import assert from "assert";
 import deepExtend from "deep-extend";
 
-export function matrixVariablesList (jobData: any, jobName: string) {
-    if (Number.isInteger(jobData.parallel)) {
+export function isPlainParallel (jobData: any) {
+    return Number.isInteger(jobData.parallel);
+}
+
+export function matrixVariablesList (jobData: any, jobName: string): {[key: string]: string}[] | null[] {
+    if (isPlainParallel(jobData)) {
         return Array(jobData.parallel).fill(null);
     }
     if (jobData?.parallel?.matrix == null) {
-        return null;
+        return [null];
     }
     assert(Array.isArray(jobData.parallel.matrix), `${jobName} parallel.matrix is not an array`);
 
--- a/src/parallel.ts
+++ b/src/parallel.ts
@@ -2,6 +2,9 @@ import assert from "assert";
 import deepExtend from "deep-extend";
 
 export function matrixVariablesList (jobData: any, jobName: string) {
+    if (Number.isInteger(jobData.parallel)) {
+        return Array(jobData.parallel).fill(null);
+    }
     if (jobData?.parallel?.matrix == null) {
         return null;
     }
--- a/tests/test-cases/parallel/integration.parallel.test.ts
+++ b/tests/test-cases/parallel/integration.parallel.test.ts
@@ -32,3 +32,16 @@ test("parallel 'test-job [1/2]'", async () => {
         chalk`{blueBright test-job: [2/2]} {greenBright >} 2/2`,
     ]));
 });
+
+test("parallel 'single-job'", async () => {
+    const writeStreams = new WriteStreamsMock();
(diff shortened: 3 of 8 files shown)


src/job.ts
```diff
@@ -169,7 +169,9 @@ export class Job {
         predefinedVariables["CI_ENVIRONMENT_NAME"] = this.environment?.name ?? "";
         predefinedVariables["CI_ENVIRONMENT_SLUG"] = this.environment?.name?.replace(/[^a-z\d]+/ig, "-").replace(/^-/, "").slice(0, 23).replace(/-$/, "").toLowerCase() ?? "";
         predefinedVariables["CI_ENVIRONMENT_URL"] = this.environment?.url ?? "";
-        predefinedVariables["CI_NODE_INDEX"] = `${opt.nodeIndex}`;
+        if (opt.nodeIndex !== null) {
+            predefinedVariables["CI_NODE_INDEX"] = `${opt.nodeIndex}`;
+        }
```
PigeonF, line 174: I tested on gitlab.com, and the variable is not set (i.e. it is not present in the environment, not set to an empty string).

Looking at the code around this though, it seems the other predefined variables are set to an empty string instead. I don't care either way, but I just wanted to point this out (since it breaks the surrounding code style).

ANGkeith:
> When writing the test I saw that there was some verification of the schema (e.g. I could not set parallel: 1), but when running gitlab-ci-local directly it works fine. Is this intended behaviour? 

probably not.

The json schema validation implemented in https://github.com/firecow/gitlab-ci-local/pull/1167 should be able to catch this.
For now, it's still hidden behind the option, `export GCL_ENABLE_JSON_SCHEMA_VALIDATION=true`

src/job.ts
ANGkeith: seems fine to me. 

But for regression testing, could you add a simple test case for this bug fix ?

src/parallel.ts
```diff
@@ -1,6 +1,28 @@
...
+        }
+    }
+
+    return {
+        matrixVariables: [null],
+        nodesTotal: null,
+    };
+}
+
 export function matrixVariablesList (jobData: any, jobName: string) {
```
ANGkeith, line 26:
```suggestion
export function matrixVariablesList (jobData: any, jobName: string) {
      if (Number.isInteger(jobData.parallel)) {
          return Array(jobData.parallel).fill(null);
      }
```

ANGkeith: let's simplify the logic like that

ANGkeith:
on a side note, @PigeonF  i think you're a collaborator right?  

do you mind pushing directly into `github.com/firecow` ?

tests/test-cases/parallel/integration.parallel.test.ts
```diff
@@ -0,0 +1,47 @@
+import {WriteStreamsMock} from "../../../src/write-streams";
+import {handler} from "../../../src/handler";
+import chalk from "chalk";
+
+test("parallel <test-job>", async () => {
+    const writeStreams = new WriteStreamsMock();
+    await handler({
+        cwd: "tests/test-cases/parallel",
+        jobs: ["test-job"],
```
ANGkeith, line 9:
```suggestion
        job: ["test-job"],
```

ANGkeith:
> Setting parallel: 1 works on https://gitlab.com/, but I cannot write a test for it because it does not match the official schema

we've copied the json schema  instead of pulling from gitlab.com (maybe we should ?),

but for now i think you can fix the schema @ https://github.com/firecow/gitlab-ci-local/blob/master/src/schema/schema.json#L1275

and your test should be good to go

PigeonF:
> on a side note, @PigeonF i think you're a collaborator right?
> 
> do you mind pushing directly into `github.com/firecow` ?

Is there some way to change the source branch of the PR directly? Or do I close this one and create a new one from the new branch?

ANGkeith:
> > on a side note, @PigeonF i think you're a collaborator right?
> > do you mind pushing directly into `github.com/firecow` ?
> 
> Is there some way to change the source branch of the PR directly? Or do I close this one and create a new one from the new branch?

for this PR it's fine, but for future PR, will be nice if you push directly to `github.com/firecow`

src/parallel.ts
```diff
@@ -1,29 +1,10 @@
...
-    return {
-        matrixVariables: [null],
-        nodesTotal: null,
-    };
-}
-
 export function matrixVariablesList (jobData: any, jobName: string) {
+    if (Number.isInteger(jobData.parallel)) {
+        return Array(jobData.parallel).fill(null);
+    }
```
PigeonF, line 7: I have implemented this simplified way like you suggested. Tell me which version you prefer and I will either drop this commit or squash it with the previous one.

ANGkeith: personally, i'll prefer this

ANGkeith (approved):

ANGkeith:
final small refactoring https://github.com/PigeonF/gitlab-ci-local/pull/1

ANGkeith (approved):

Issue #1233, opened 2024-05-26
Cannot override CLI args that have default values using `.gitlab-ci-local-env` file


PigeonF:
**Minimal .gitlab-ci.yml illustrating the issue**

```yml
---
job:
  image: docker.io/busybox
  script:
    - echo "Hello" > hello.txt
  artifacts:
    paths:
      - hello.txt
```

Compare the following ways to run `gitlab-ci-local`

1. Using a CLI arg (this works as expected)

```console
$ gitlab-ci-local --no-artifacts-to-source
$ ls hello.txt
"hello.txt": No such file or directory
```

2. Using environment variable override (this works as expected)

```console
$ GCL_ARTIFACTS_TO_SOURCE=false gitlab-ci-local
$ ls hello.txt
"hello.txt": No such file or directory
```

3. Using an environment variable override in `.gitlab-ci-local-env` (this does not work)

```console
$ cat .gitlab-ci-local-env
ARTIFACTS_TO_SOURCE=false
$ gitlab-ci-local
$ ls hello.txt
hello.txt
```

As far as I can tell, the issue is that the command line option gets a default value assigned in `index.ts`

https://github.com/firecow/gitlab-ci-local/blob/d44cedb180f718c4af09870809124d4b44546b03/src/index.ts#L213-L218

and then in `argv.ts` the value is only overriden if it was undefined before

https://github.com/firecow/gitlab-ci-local/blob/d44cedb180f718c4af09870809124d4b44546b03/src/argv.ts#L71-L73

This should apply to all command line options with a default value.

firecow:
Yup, we cannot use yargs default values, when doing override like we do.

PR #1234, opened 2024-05-26
Remove defaults from yargs, they cannot be overridden by other GCL mechanism

firecow:
- Remove the enable part from enableJsonSchemaValidation CLI option, no other bool CLI options has the word enable so it seems redundant

- Replace line count in all tests other than plain  (less assertion roulette)

- Replace dnsutils with a maintained one

closes https://github.com/firecow/gitlab-ci-local/issues/1233

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -254,8 +254,8 @@ export class Argv {
         return this.map.get("containerExecutable") ?? "docker";
     }
 
-    get enableJsonSchemaValidation (): boolean {
-        return this.map.get("enableJsonSchemaValidation") ?? false;
+    get jsonSchemaValidation (): boolean {
+        return this.map.get("jsonSchemaValidation") ?? false;
     }
 
     get shellExecutorNoImage (): boolean {
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -255,7 +255,7 @@ export class Argv {
     }
 
     get enableJsonSchemaValidation (): boolean {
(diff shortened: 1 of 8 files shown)


src/argv.ts
```diff
@@ -254,8 +256,9 @@ export class Argv {
         return this.map.get("containerExecutable") ?? "docker";
     }
 
-    get enableJsonSchemaValidation (): boolean {
-        return this.map.get("enableJsonSchemaValidation") ?? true;
+    get jsonSchemaValidation (): boolean {
```
ANGkeith, line 259: This might be breaking changes for people that's using `enableJsonSchemaValidation`.

We might want to emphasis this in the release note of this breaking changes in this "beta" flag

also, agree that `enable` is redundant here

src/argv.ts
```diff
@@ -225,6 +226,7 @@ export class Argv {
     }
 
     get artifactsToSource (): boolean {
+        // TODO: default to false in 5.x.x
```
ANGkeith, line 229: Is this a typo?

src/argv.ts
firecow: Yes, it's acceptable, since the flag is so new.

src/argv.ts
firecow: artifacts to source will be default "off" in 5.x.x

ANGkeith (approved):

PR #1248, opened 2024-06-03
Do not validate formats and remove all string patterns from schema.json

firecow:
```yml
---
test:
  environment:
    name: test
    url: ''
  script: hello
```

This make gitlab-ci-local go boom.

But gitlab is fine with it :smile: Which leaves me to believe that they aren't enforcing their json schema string patterns or formats, so let's ignore them for now.

### DIFF
--- a/src/validator.ts
+++ b/src/validator.ts
@@ -3,7 +3,6 @@ import {Job} from "./job";
 import assert from "assert";
 import chalk from "chalk";
 import * as yaml from "js-yaml";
-import addFormats from "ajv-formats";
 import schema from "./schema";
 
 export class Validator {
@@ -12,17 +11,13 @@ export class Validator {
             verbose: true,
             allErrors: true,
             allowUnionTypes: true,
+            validateFormats: false,
             strictTypes: false, // to suppress the missing types defined in the gitlab-ci json schema
+            keywords: ["markdownDescription"],
         });
-        ajv.addKeyword("markdownDescription");
-        addFormats(ajv);
         const validate = ajv.compile(schema);
         const valid = validate(data);
-        assert(valid,
-            chalk`Invalid gitlab-ci configuration! It have failed the json schema validation. Dump the following to the pipeline editor to debug:
-
-${yaml.dump(data)}
-`);
+        assert(valid, chalk`Invalid gitlab-ci configuration! It have failed the json schema validation. Dump the following to the pipeline editor to debug: ${yaml.dump(data)}`);
     }
 
     private static needs (jobs: ReadonlyArray<Job>, stages: readonly string[]): string[] {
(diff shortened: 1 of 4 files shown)


ANGkeith:
I see that they’ve accept @PigeonF mr. https://gitlab.com/gitlab-org/gitlab/-/issues/462979#note_1923462655

> Our developers found many issues today with this approach, a lot of local pipelines was failing.


is there a lot of error with the json schema? Or is it worth the trouble to fix the schema and contribute to gitlab.com

src/validator.ts
```diff
@@ -12,17 +11,13 @@ export class Validator {
...
-        ajv.addKeyword("markdownDescription");
-        addFormats(ajv);
         const validate = ajv.compile(schema);
         const valid = validate(data);
-        assert(valid,
-            chalk`Invalid gitlab-ci configuration! It have failed the json schema validation. Dump the following to the pipeline editor to debug:
-
-${yaml.dump(data)}
-`);
+        assert(valid, chalk`Invalid gitlab-ci configuration! It have failed the json schema validation. Dump the following to the pipeline editor to debug: ${yaml.dump(data)}`);
```
ANGkeith, line 20: I see the final new line have been removed. 

I’m away from computer to test it locally, but does  yaml.dump already have a new line ?

ANGkeith (approved):

firecow:
> I see that they’ve accept @PigeonF mr. https://gitlab.com/gitlab-org/gitlab/-/issues/462979#note_1923462655
> 
> > Our developers found many issues today with this approach, a lot of local pipelines was failing.
> 
> is there a lot of error with the json schema? Or is it worth the trouble to fix our schema and contribute upstream to gitlab.com

After some digging around, its only the string pattern and formats that weren't being upheld by Gitlab CI/CD, so those were the only issues we ran into.

We cannot contribute to upstream Gitlab.com on how they interpret their json schema in Gitlab CI/CD can we?

ANGkeith:
> We cannot contribute to upstream Gitlab.com on how they interpret their json schema in Gitlab CI/CD can we?

Not sure, maybe we can create a “feature request” for it. 

There might be some reasons / limitation why they did not conduct the schema validation on the backend😅
But looking at the current state, there’s still some tuning that needs to be done in the json schema specification otherwise it’ll result in breaking changes to current user

> contribute upstream to gitlab.com
I am referring to contributing to the json schema as per what pigeonf done

firecow:
Aaah, well.. I haven't found any mistakes in the json schema definition, though.
If we find any errors, lets contribute to the schema.

PigeonF:
> There might be some reasons / limitation why they did not conduct the schema validation on the backend😅

I believe they do not do any schema validation on the backend - they do the validation in the ruby code ([e.g. here](https://gitlab.com/gitlab-org/gitlab/-/blob/d1157576ce6fbb0f54e398e82d168cbc9c6cec22/lib/gitlab/ci/config/entry/product/parallel.rb#L21-23)). As far as I can tell, the schema is only used for the pipeline editor web UI.

PR #1253, opened 2024-06-06
early return from rules evaluation after first match

### LINKED ISSUE #1252: Rules evaluation is incorrect and it doesn't stop after the first match
```yml
---
variables:
  VAR1: val1
  VAR2: val2

Test:
  script:
    - echo "hello"
  rules:
    - if: $VAR1 == 'val1'
    - if: $VAR2 == 'val2'
      when: never
```

Running `gitlab-ci-local --list` with the above gitlab ci configuration produces incorrect results (it doesn't list the Test job)
```
parsing and downloads finished in 63 ms
json schema validated in 95 ms
name  description  stage   when   allow_failure  needs
```

**Expected behavior**
`gitlab-ci-local --list` should list the "Test" job as the first rule evaluates to true.

**Host information**
Mac 14.4.1
gitlab-ci-local 4.50.0 and above

**Additional context**
This started happening with version 4.50.0. Possibly due to https://github.com/firecow/gitlab-ci-local/commit/ef8097bedca05230bef15a3f52d395a6414eef23 as it doesn't stop evaluating rules unless when = never

But [Gitlab docs for rules](https://docs.gitlab.com/ee/ci/yaml/#rules) says

> Rules are evaluated when the pipeline is created, and evaluated in order until the first match

vivek9ru3:
fixes #1252

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -196,7 +196,7 @@ export class Utils {
             allowFailure = rule.allow_failure ?? allowFailure;
             ruleVariable = rule.variables;
 
-            if (when === "never") break; // Early return, will not evaluate the remaining rules
+            break; // Early return, will not evaluate the remaining rules
         }
 
         return {when, allowFailure, variables: ruleVariable};
--- a/tests/rules.test.ts
+++ b/tests/rules.test.ts
@@ -299,3 +299,13 @@ test("optional manual job", () => {
     const rulesResult = Utils.getRulesResult({cwd: "", rules, variables}, gitData, jobWhen);
     expect(rulesResult).toEqual({when: "manual", allowFailure: true, variables: undefined});
 });
+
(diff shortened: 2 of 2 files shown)


ANGkeith (approved):

src/utils.ts
```diff
@@ -196,7 +196,7 @@ export class Utils {
             allowFailure = rule.allow_failure ?? allowFailure;
             ruleVariable = rule.variables;
 
-            if (when === "never") break; // Early return, will not evaluate the remaining rules
+            break; // Early return, will not evaluate the remaining rules
```
ANGkeith, line 199:
```suggestion
            break; // Rules are evaluated in order until the first match
```

ANGkeith:
LGTM! thanks for the fix

ANGkeith:
fixes #1252

PR #1257, opened 2024-06-06
refactor: make maxiumIncludes configurable

ANGkeith:


ANGkeith:
> Can we just print a warning instead?
Currently, this also act as a safety net to breakout of circular includes.

>  I don't think the check here is accurate, as gitlab-ci-local fails to run a pipeline which works in our GitLab instance (at least it might count includes a bit differently?

could you try to hack gitlab-ci-local src code to show whats the `MAXIMUM_INCLUDES` that gitlab-ci-local calculated?

hverlin (approved):

Issue #1255, opened 2024-06-06
Make validation optional


hverlin:
**Is your feature request related to a problem? Please describe.**
Having the validation check using the schema is nice. However, sometimes it fails to run a "valid" (as is "it runs on GitLab") pipeline.

I believe it might also be an issue if you are using a private GitLab instance which might not support the latest version of the schema?

**Describe the solution you'd like**
Make all validation optional, or add an option to print a warning instead?

**Describe alternatives you've considered**
Version later than [4.46.1](https://github.com/firecow/gitlab-ci-local/releases/tag/4.46.1) fails because of invalid schema or max. includes check

The workaround for me is to manually update the code or use 
```
npm i -g gitlab-ci-local@4.46.0
```

ANGkeith:
> Having the validation check using the schema is nice. However, sometimes it fails to run a "valid" (as is "it runs on GitLab") pipeline.


yeah unfortunately because gitlab seem to perform the schema validation on their backend
https://github.com/firecow/gitlab-ci-local/pull/1248#issuecomment-2146971764

to disable the schema validation 
`export GCL_JSON_SCHEMA_VALIDATION='false'`

ANGkeith:
> (Also, hard-coding it to 150 is not right. Maybe it should come from an env. variable as it can be configured in the GitLab admin) https://github.com/firecow/gitlab-ci-local/pull/1123#issuecomment-2152192377 
Totally agree on this

ANGkeith:
@hverlin regarding https://github.com/firecow/gitlab-ci-local/pull/1123#issuecomment-2152192377

> Can we just print a warning instead?

Currently, this also act as a safety net to breakout of circular includes.

>  I don't think the check here is accurate, as gitlab-ci-local fails to run a pipeline which works in our GitLab instance (at least it might count includes a bit differently?

could you try to hack gitlab-ci-local src code to show whats the `MAXIMUM_INCLUDES` that `gitlab-ci-local` calculated?

hverlin:
> could you try to hack gitlab-ci-local src code to show whats the MAXIMUM_INCLUDES that gitlab-ci-local calculated?
@ANGkeith I am getting 156 includes :sweat_smile:

hverlin:
Other suggestion: maybe printing all the includes on error would help. In our case, it made pretty obvious which includes were included several times.

```ts
export class ParserIncludes {
  private static count: number = 0;
  private static allIncludes: Array<string> = [];

[...]

PR #1266, opened 2024-06-16
Improve validation #1263 #1255

### LINKED ISSUE #1263: Revisit if `types` is indeed an illegal name
```yml
---
types:
  script:
    - echo "test"

test:
  needs: [types]
  script:
    - echo "test"
```

**Expected behavior**
According to the documentation, it should not work.
https://docs.gitlab.com/ee/ci/jobs/#job-name-limitations

However, it does work:
https://gitlab.com/hverlin/gitlab-ci-local-tests/-/pipelines/1324968259

Probably this is a limitation that they forgot to remove from the documentation?

```shell
> gitlab-ci-local
needs: [types] for test could not be found
```

Two solutions:
- remove `types` from: https://github.com/firecow/gitlab-ci-local/blob/ba3509a42b5e0155913fb09d0885c93284df1f39/src/job.ts#L69-L73
- add a warning to the console if a job name matches an invalid name

hverlin:
Fix #1263 #1255

Given 
```yml
test:
  needs: [types]
  when: 'manual2'
  artifacts:
    reports:
      junit: 3
  script: ''
  data: "test"
```

```
Invalid .gitlab-ci.yml configuration
        • 'data' property is not expected to be here at test
        • property 'script' must not have fewer than 1 characters at test.script
        • 'when' property must be one of [on_success, on_failure, always, never, manual, delayed] (found manual2) at test.when
        • 'junit' property type must be string at test.artifacts.reports.junit
```
image


hverlin:
@firecow @ANGkeith Let me know if these changes are ok. I will then add tests.

- Regarding schema validation, output was done directly using `console`/`assert` before. Should I switch to `writeStreams`?
- I am now always writing the expanded pipeline `yaml` under the `.gitlab-ci-local`. I think this is useful outside validation in case you want to inspect what was parsed without having to always do `gcl --preview > /tmp/my-pipeline.yml`

ANGkeith:
looks good to me

hverlin:
@firecow Anything you would change here before I add tests?

firecow:
@hverlin Nah, nothing much. All looks good.


```
Validation errors:
        test must NOT have additional properties
```

Is it possible to print exactly what properties are additional?

hverlin:
I reused code from https://github.com/apideck-libraries/better-ajv-errors, and I am now getting better error message.

Given 
```yml
test:
  needs: [types]
  when: 'manual2'
  artifacts:
    reports:
      junit: 3
  script: ''
  data: "test"
```

```
Invalid .gitlab-ci.yml configuration
        • 'data' property is not expected to be here at test
        • property 'script' must not have fewer than 1 characters at test.script
        • 'when' property must be one of [on_success, on_failure, always, never, manual, delayed] (found manual2) at test.when
        • 'junit' property type must be string at test.artifacts.reports.junit
```
image

src/index.ts
```diff
@@ -30,7 +30,7 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
         .parserConfiguration({"greedy-arrays": false})
         .showHelpOnFail(false)
         .version(packageJson["version"])
-        .wrap(yargs.terminalWidth())
+        .wrap(yargs.terminalWidth?.())
```
hverlin, line 33: This makes it possible to run  `gitlab-ci-local` with `bun` to that it's faster to iterate
```
bun ../../gitlab-ci-local/src/index.ts
```

src/schema-error.ts
```diff
@@ -0,0 +1,142 @@
+// adapted from https://github.com/apideck-libraries/better-ajv-errors (MIT)
```
hverlin, line 1: I did not use the code directly as:
- I was getting NPE for some cases
- I am not sure if it makes sense to bring the other libraries that they use for suggestions
- We can improve some error messages later if needed

firecow: I don't particularly like this, but I'll let it slip :smile:

firecow (approved):

firecow:
@ANGkeith I'll let you review this as well, before I'm merging.

ANGkeith:
very nice

src/validator.ts
```diff
@@ -16,8 +23,27 @@ export class Validator {
...
+                errors: validate.errors,
+            });
+
+            const errors = betterErrors.map((e => (chalk`\t• {redBright ${e.message}} at {blueBright ${e.path}}`)));
+            writeStreams.stderr(errors.splice(0, MAX_ERRORS).join("\n"));
+            if (errors.length > MAX_ERRORS) {
+                writeStreams.stderr(`\t... and ${errors.length - MAX_ERRORS} more`);
+            }
+
+            writeStreams.stderr(chalk`\nWhat to do next:
```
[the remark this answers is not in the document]
hverlin: Updated it with link support
image

ANGkeith (approved):

PR #1294, opened 2024-07-13
fix: Allow repeated include inputs interpolation

PigeonF:
A template like

```yml
---
spec:
  inputs:
    name:
---
job $[[ inputs.name ]] $[[ inputs.name ]]:
  script:
    - echo $[[ inputs.name ]]$[[ inputs.name ]]
```

currently ends up looking like

```yaml
job name $[[ inputs.name ]]:
  script:
    - echo name$[[ inputs.name ]]
```

after substitution.

The fix is to make the regex parsing the `$[[ ]]` part more permissive.

### DIFF
--- a/src/parser.ts
+++ b/src/parser.ts
@@ -308,7 +308,7 @@ export class Parser {
 
             const interpolatedConfigurations = JSON.stringify(uninterpolatedConfigurations)
                 .replace(
-                    /(?<firstChar>.)(?<secondChar>.)\$\[\[\s*inputs.(?<interpolationKey>[\w-]+)\s*\|?\s*(?<interpolationFunctions>.*?)\s*\]\](?<lastChar>.)/g // https://regexr.com/81c16
+                    /(?<firstChar>.)?(?<secondChar>.)?\$\[\[\s*inputs.(?<interpolationKey>[\w-]+)\s*\|?\s*(?<interpolationFunctions>.*?)\s*\]\](?<lastChar>[^$])?/g // https://regexr.com/81c16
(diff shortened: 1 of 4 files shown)


src/parser.ts
```diff
@@ -308,7 +308,7 @@ export class Parser {
 
             const interpolatedConfigurations = JSON.stringify(uninterpolatedConfigurations)
                 .replace(
-                    /(?<firstChar>.)(?<secondChar>.)\$\[\[\s*inputs.(?<interpolationKey>[\w-]+)\s*\|?\s*(?<interpolationFunctions>.*?)\s*\]\](?<lastChar>.)/g // https://regexr.com/81c16
+                    /(?<firstChar>.)?(?<secondChar>.)?\$\[\[\s*inputs.(?<interpolationKey>[\w-]+)\s*\|?\s*(?<interpolationFunctions>.*?)\s*\]\](?<lastChar>[^$])?/g // https://regexr.com/81c16
```
PigeonF, line 311: We have to make the first two chars optional, because the string might have already been consumed up to the `$[[` by a previous iteration of the `replace` call.

The last char has to exclude the `$` so that the current iteration of the `replace` does not consume the beginning of what is supposed to be part of the next iteration. This means that the current iteration might not have a `lastChar` if it is immediately followed by the `$[[` of the next iteration, so `lastChar` has to be optional as well.

PigeonF: Also, I don't know how to update the regexr link 😅

ANGkeith: > Also, I don't know how to update the regexr link 😅

updated! it's under my account 😅

ANGkeith (approved):
lgtm. Thanks for spotting and fixing

PR #1295, opened 2024-07-14
Expand variables in `rules:changes`

PigeonF:
This enables common `parallel:matrix` configurations like

```yaml
matrix:
  image: alpine
  parallel:
    matrix:
      - SUBDIR:
          - foo
          - bar
  rules:
    - changes:
        paths:
          - "${SUBDIR}/**/*"
  script:
    - echo "Job is running"
```

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -79,11 +79,12 @@ export class Job {
     readonly environment?: {name: string; url: string | null; deployment_tier: string | null; action: string | null};
     readonly jobId: number;
     readonly rules?: {
-        if: string;
-        when: string;
-        exists: string[];
-        allow_failure: boolean;
-        variables: {[key: string]: string};
+        if?: string;
+        when?: string;
+        changes?: string[] | {paths: string[]};
+        exists?: string[];
+        allow_failure?: boolean;
+        variables?: {[name: string]: string};
     }[];
 
     readonly allowFailure: boolean | {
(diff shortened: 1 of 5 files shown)


src/job.ts
```diff
@@ -79,11 +79,12 @@ export class Job {
...
-        when: string;
-        exists: string[];
-        allow_failure: boolean;
-        variables: {[key: string]: string};
+        if?: string;
+        when?: string;
+        changes?: string[] | {paths: string[]};
+        exists?: string[];
+        allow_failure?: boolean;
+        variables?: {[name: string]: string};
```
PigeonF, line 87: I changed this to match https://github.com/firecow/gitlab-ci-local/blob/70d2462f8b00fb74e902ca33ed87cdcaee6b083a/src/utils.ts#L16-L23

src/job.ts
```diff
@@ -181,6 +182,26 @@ export class Job {
...
+                if (!changes || !Array.isArray(changes)) {
+                    return;
+                }
+
+                changes.forEach((change, changeIdx) => {
+                    changes[changeIdx] = Utils.expandText(change, expanded);
+                });
+
+                assert(this.rules);
+                assert(this.rules[ruleIdx]);
```
PigeonF, line 199: I'm not sure why the typescript compiler thinks `this.rules` or `this.rules[ruleIdx]` can be undefined at this point, but adding the asserts is required for the assignment below.

src/job.ts
```diff
@@ -181,6 +182,26 @@ export class Job {
...
+
+                this.rules[ruleIdx].changes = changes;
+            });
+        }
+
```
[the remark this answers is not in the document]
PigeonF: Thanks, I did not know about the third parameter to `forEach` 👍

src/job.ts
ANGkeith: lets export this 

ie. 

```typescript
export type JobRule = {
    if?: string;
    when?: string;
    changes?: string[] | {paths: string[]};
    exists?: string[];
    allow_failure?: boolean;
    variables?: {[name: string]: string};
};
```
then import it in src/utils

so that we don't need to sync the types in two places

src/job.ts
```diff
@@ -181,6 +182,21 @@ export class Job {
         predefinedVariables["CI_REGISTRY"] = `local-registry.${this.gitData.remote.host}`;
         predefinedVariables["CI_REGISTRY_IMAGE"] = `$CI_REGISTRY/${this._variables["CI_PROJECT_PATH"].toLowerCase()}`;
 
+        // Expand variables in rules:changes
+        if (this.rules && expandVariables) {
+            const expanded = Utils.expandVariables(this._variables);
+            this.rules.forEach((rule, ruleIdx, rules) => {
+                const changes = Array.isArray(rule.changes) ? rule.changes : rule.changes?.paths;
+                if (!changes || !Array.isArray(changes)) {
```
ANGkeith, line 190:
```suggestion
                if (!changes) {
```
what is `!Array.isArray(changes)` for?

PigeonF: To guard against `rule.changes?.paths` not being an array.

ANGkeith: > since schema validation can be disabled
In my opinion, i think this can be removed,

Because on a typical(99% of users) scenario, schema validation should never be disabled, unless there's some bug with it. 

otherwise these type of defensive statement will eventually bloat this program in the future.

ANGkeith (approved):

PR #1315, opened 2024-08-15
fix: Allow multi-line variable values via command line

daniel-albuschat:
When passing variables via command line via --variable, and the variable had multi-line content, only the first line was passed onto the container.

Example:

```bash
gitlab-ci-local --variable MULTILINE="This is
a multi line
variable value"
```

In this case, only "This is" was passed to the CI job.


daniel-albuschat:
@firecow Thanks for starting the checks! 👍 I will fix the eslint error. However, I don't think that the failing jest tests are related to my changes. Looks like the tests might be flaky?

firecow (approved):

daniel-albuschat:
@firecow Tak skal du ha'! 👍

PR #1355, opened 2024-10-02
feat: customizable project file variable file path and support dotenv format

### LINKED ISSUE #1332: Add --env-file-to-variables cli option
 Currently, I must store the .env file in one directory, limiting flexibility, and having to manager two separate .env files. I'd like to be able to specify the .env file's location via an environment variable so it can be used from any directory or path.

add the option to define the path to the .env file using an environment variable, like this: export GCL_ENV_FILE=.env. This would allow the gitlab-ci-local command to use the .env file from any given path, regardless of its directory. This change would allow for more flexibility in working with multiple environments and project structures.

ANGkeith:
resolves #1332


src/index.ts
```diff
@@ -109,6 +109,12 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
             description: "Path to a current working directory",
             requiresArg: true,
         })
+        .option("local-variables-file", {
```
ANGkeith, line 112: naming is hard, let me know if you want to name this something else

src/index.ts
```diff
@@ -109,6 +109,12 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
...
+            type: "string",
+            description: "Path to the project file variables",
+            requiresArg: false,
+            default: Argv.default.localVariablesFile,
```
ANGkeith, line 116: introducing this new pattern, so that `--help` will show the default values which i find quite useful..

image

if we're ok with this, will open a new mr to introduce same pattern for the other `options`

src/index.ts
firecow: `gitlab-ci-local --variables-file=/home/mjn/something-custom/variables.yml`

No need to emphasize that its local, you are running `gitlab-ci-LOCAL` already.

firecow:
I don't like this complexity... 
It's a different file format, it should be a different command line option.

ANGkeith:
> I don't like this complexity...
It's a different file format, it should be a different command line option.

I don't really see why this increase the complexity though.. it's different file format, but it serves the same purpose (ie. setting variables for the executor)


taking `eslint` as an example, they don't create --json-config --mjs-config ... you get my point.

firecow:
> > I don't like this complexity...
> > It's a different file format, it should be a different command line option.
> 
> I don't quite understand your perspective on this.
> 
> The way I see it is: it's different file format, but it serves the same purpose (ie. setting variables for the executor)
> 
> Taking `eslint` as an example: It only has a single `--config` option, regardless of the supported format. They don’t introduce separate options for each format (e.g., --json-config, --mjs-config, etc.).

You convinced me with the eslint example. Get it out of draft, and lets merge it :+1:

ANGkeith:
nice! 

let's get the big mr https://github.com/firecow/gitlab-ci-local/pull/1331 in first.

firecow (approved):

PR #1360, opened 2024-10-03
feat: add support for absolute path for options that makes sense

### LINKED ISSUE #1067: Decouple from cwd

I would like to keep all gitlab-ci-local related data outside of my git projects, completely separate. Which seems to be an impossible goal due to the common pattern to bind almost all path operations to the `cwd` (`${cwd}/${stateDir}/...`, `${cwd}/${file}`, git, etc.).

I would like to have an option to be able to decouple from the `cwd`. The idea here is that GCL should take only a source code from the particular directory I could have specified, while everything else, like dotfiles, variables.yaml, stateDir, even gitlab-ci-local.yml itself, sits in another directory, outside of a git project. Something like `--git-dir` CLI option, maybe.

I was trying to combine different CLI options (`--home`, `--state-dir`, `--cwd`, `--file`) in attempt to achieve what I want, but I was unable to do so.

**Additional context**

N/A

ANGkeith:
resolves #1067

# breaking changes
- `--home` no longer supports relative path, since i don't think it make sense.
  
-


ANGkeith:
possible to make a major release soon ? 

This mr touches too much file.. don't really want to leave it lying around for too long haha

firecow:
Closing - this draft has been open since October 2024 with no reviews. It also has some bugs (stray brace in home getter, rsyncTrackedFiles assumes relative stateDir) and breaking changes that need migration guidance. Can be revisited if there's renewed interest.

PR #1368, opened 2024-10-08
fix: better support for custom gitlab ports

ANGkeith:

PR #1369, opened 2024-10-09
Fix/better support for custom gitlab ports

rgalonso:
This includes changes to properly support a self-hosted instance with non-standard host ports (and schema).

This PR also includes changes to support the [Dependency Proxy](https://docs.gitlab.com/ee/user/packages/dependency_proxy/). If you'd rather keep that separated as a separate PR, let me know. I've kept the commits separate to easily isolate these items.


ANGkeith:
>  If you'd rather keep that separated as a separate PR, let me know. I've kept the commits separate to easily isolate these items.
yup

Nice work, will be nice to keep them as a seperate PR. 

Would you like to try adding some tests?

src/predefined-variables.ts
```diff
@@ -13,16 +13,27 @@ export function init ({gitData, argv, envMatchedVariables}: PredefinedVariablesO
     // 1. gitlab variables files
     // 2. values derieved implicitly from `git remote -v`
     // 3. default value
+    const CI_SERVER_PROTOCOL = envMatchedVariables["CI_SERVER_PROTOCOL"]
```
ANGkeith, line 16: did not add this in.. 

it's 2024, i wonder who still uses `http` HAHA

ANGkeith, line 16: It's 2024, i wonder who still uses http HAHA

anyways, this is still nice to have

src/predefined-variables.ts
```diff
@@ -13,16 +13,27 @@ export function init ({gitData, argv, envMatchedVariables}: PredefinedVariablesO
...
+        : "443");
     const CI_SERVER_SHELL_SSH_PORT = envMatchedVariables["CI_SERVER_SHELL_SSH_PORT"]
-      ?? (gitData.remote.schema === "ssh")
+      ?? ((gitData.remote.schema === "ssh")
         ? gitData.remote.port
-        : "22";
+        : "22");
+    const CI_SERVER_HOST = envMatchedVariables["CI_SERVER_HOST"]
+      ?? `${gitData.remote.host}`;
+    const CI_SERVER_FQDN = envMatchedVariables["CI_SERVER_FQDN"]
```
ANGkeith, line 30: const CI_SERVER_FQDN = CI_SERVER_PORT == "443" ? gitData.remote.host : `${gitData.remote.host}:${CI_SERVER_PORT}`;

to be consistent with gitlab.com... the port for the FQDN is implicit when it's 443 

as seen github.com/firecow/gitlab-ci-local/pull/1368#issuecomment-2398753437

src/predefined-variables.ts
```diff
@@ -13,16 +13,29 @@ export function init ({gitData, argv, envMatchedVariables}: PredefinedVariablesO
...
+      ?? `${CI_SERVER_HOST}:${CI_SERVER_PORT}`;
```
ANGkeith, line 31:
```suggestion
    const CI_SERVER_FQDN = envMatchedVariables["CI_SERVER_FQDN"]
      ?? (CI_SERVER_PORT == "443"
          ? gitData.remote.host
          : `${gitData.remote.host}:${CI_SERVER_PORT}`);
```
to be consistent with gitlab.com... the port for the FQDN is implicit when it's 443 

as seen https://github.com/firecow/gitlab-ci-local/pull/1368#issuecomment-2398753437

src/predefined-variables.ts
```diff
@@ -50,16 +64,20 @@ export function init ({gitData, argv, envMatchedVariables}: PredefinedVariablesO
...
-        CI_SERVER_HOST: `${gitData.remote.host}`,
+        CI_SERVER_FQDN: CI_SERVER_FQDN,
+        CI_SERVER_HOST: CI_SERVER_HOST,
         CI_SERVER_PORT: CI_SERVER_PORT,
         CI_SERVER_SHELL_SSH_PORT: CI_SERVER_SHELL_SSH_PORT,
         CI_SERVER_URL: CI_SERVER_URL,
-        CI_SERVER_PROTOCOL: "https",
+        CI_SERVER_PROTOCOL: CI_SERVER_PROTOCOL,
         CI_API_V4_URL: `${CI_SERVER_URL}/api/v4`,
         CI_PROJECT_URL: `${CI_SERVER_URL}/${gitData.remote.group}/${gitData.remote.project}`,
```
ANGkeith, line 74:
```suggestion
        CI_PROJECT_URL: `${CI_SERVER_FQDN}/${gitData.remote.group}/${gitData.remote.project}`,
```

ANGkeith, line 74:
```suggestion
        CI_API_V4_URL: `${CI_SERVER_FQDN}/api/v4`,
        CI_PROJECT_URL: `${CI_SERVER_FQDN}/${gitData.remote.group}/${gitData.remote.project}`,
```

rgalonso:
> > If you'd rather keep that separated as a separate PR, let me know. I've kept the commits separate to easily isolate these items.
> 
> yup, will be nice to keep them as a seperate PR. for easier review
>
OK, I'll re-push with just the first commit included.
 
> Nice work ! and even nicer, you can test that indeed these changes works for your self hosted gitlab use-case.. saves us from several back and forth
Thanks! It was fun learning a new language.

> 
> Is it very late on your side ? Would you like to try adding some tests? if not i can add them, i have some free time
Yes, it is kind of late here. I'm Eastern Time US. If you have the time, I'd appreciate it and then I can test it out tomorrow. Also, since I'm new to TypeScript, I'm not really sure of the idiomatic way to write tests. I saw that you had written a test on the original branch (`feat/add-dependency-proxy-variables`), but then didn't include that in your `fix/better-support-for-custom-gitlab-ports`. Was that because it wasn't working, or just because you figured it was incomplete since you can't test it directly?

src/predefined-variables.ts
rgalonso: We really do want `CI_SERVER_URL` here. That has the schema prefix (`https://`), whereas `CI_SERVER_FQDN` does not. Applying the suggested change would be incompatible with GitLab and not be backwards compatible.

ANGkeith: opps, yes you're right

src/predefined-variables.ts
rgalonso: I know, right? Even in my internal instance, HTTPS is used. That said, HTTP was previously used when the server was completely isolated from the Internet. So as long as GitLab continues to support it, so should we.

ANGkeith:
> I saw that you had written a test on the original branch (feat/add-dependency-proxy-variables), but then didn't include that in your fix/better-support-for-custom-gitlab-ports

ahh because those were testing for the depedency proxy variables related stuff


and my `fix/better-support-for-custom-gitlab-ports` mr was more for fixing the predefined variables for custom gitlab ports users

rgalonso:
> > I saw that you had written a test on the original branch (feat/add-dependency-proxy-variables), but then didn't include that in your fix/better-support-for-custom-gitlab-ports
> 
> ahh because those were testing for the depedency proxy variables related stuff
> 
> and my `fix/better-support-for-custom-gitlab-ports` mr was more for fixing the predefined variables for custom gitlab ports users

Ah, right. With all the back and forth, I forgot that the test was just for that feature.

rgalonso:
I see some of the regression [tests failed](https://github.com/firecow/gitlab-ci-local/actions/runs/11247410323/job/31270874491). Sorry about that. Is there a way to test that locally? (Again, new to TypeScript.)

ANGkeith:
no worries, most of it are rather trivial.. 

> Is there a way to test that locally?
https://github.com/firecow/gitlab-ci-local?tab=readme-ov-file#scripts

have you seen ^, are you facing any weird errors ?

rgalonso:
I'd forgotten that was in the README. I was looking for a separate CONTRIBUTING.md or the like. That's working for me now. Thanks.

PR #1370, opened 2024-10-09
fix linting/regression test issues

rgalonso:


ANGkeith:
i've cherry picked your commits

thanks

PR #1375, opened 2024-10-13
fix: additional '>' char generated in command shell prompt when gcl --no-color

ANGkeith:


src/job.ts
```diff
@@ -860,7 +861,11 @@ export class Job {
         const outFunc = (line: string, stream: (txt: string) => void, colorize: (str: string) => string) => {
             this.refreshLongRunningSilentTimeout(writeStreams);
             stream(`${this.formattedJobName} `);
-            if (!line.startsWith("\u001b[32m$")) {
```
ANGkeith, line 863: `if (!line.startsWith("\u001b[32m$")) {`


when --no-color, this condition will never be true

firecow (approved):

PR #1377, opened 2024-10-14
fix: wrong line number in stack trace

ANGkeith:
npm uninstall ts-node source-map-support @types/source-map-support
npm install --save-dev tsx

# Issue
Using the following 
```yaml
# .gitlab-ci.yaml
---
job:
  image: test-err
  script:
    - echo 1
```

image
The stack trace returns the wrong line number. 

which is possibly related to https://www.github.com/TypeStrong/ts-node/issues/2053, thus replacing `ts-node` with `tsx`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,6 @@
         "object-traversal": "1.x.x",
         "p-map": "4.x.x",
         "pretty-hrtime": "1.x.x",
-        "source-map-support": "0.5.x",
         "split2": "4.x.x",
         "terminal-link": "2.1.1",
         "yargs": "17.x.x"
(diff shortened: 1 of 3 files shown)


firecow (approved):

firecow:
merge with master please @ANGkeith

firecow:
@ANGkeith We need to ensure that stack traces in the pkg binaries still works, when you remove the sourcemap package.

ANGkeith:
opps, forgot about that..

but do we really need sourcemap for the binary ? 

i thought sourcemap is more for dev/debugging

ANGkeith:
anyways, 

> @ANGkeith We need to ensure that stack traces in the pkg binaries still works, when you remove the sourcemap package.

should be fixed already

firecow:
@ANGkeith We don't have any security concerns going on in this tool, so I can't think of a reason why we would want to exclude such epic piece of debug information, when users make bug reports.

PR #1385, opened 2024-10-18
refactor: remove unnecessary repeated `reference` call

ANGkeith:


firecow (approved):
Woaw, nice catch. I spent ages getting that right.

PR #1397, opened 2024-10-24
Added --default-image option to be used with --shell-executor-no-image=false

NGPetrov:
Currently if `--shell-executor-no-image=false` is used the runner defaults to `docker.io/ruby:3.1`.
In case isolation is required, there is no option to run the ci in a custom image.

Added --default-image option. 
If set when `--shell-executor-no-image=false`, the value will be used as a image source for the container.
If not set when `--shell-executor-no-image=false`, the default `docker.io/ruby:3.1` will be used.
If set when `--shell-executor-no-image=true` (or not set), warring message is displayed.
If set when `--shell-isolation`, warring message is displayed.

~Fixed typo (or old option name) in `arvg.ts:49`, from `--no-shell-executor-no-image` to  `--shell-executor-no-image`~

Included 3 tests to validate proper runner image/shell is used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "depcheck": "1.x.x",
         "esbuild": "^0.24.0",
         "eslint": "8.x.x",
-        "jest": "^29.7.0",
+        "jest": "29.x.x",
         "jest-when": "3.x.x",
         "nodemon": "^3.1.1",
         "ts-jest": "29.x.x",
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,7 +53,7 @@
         "depcheck": "1.x.x",
         "esbuild": "^0.24.0",
         "eslint": "8.x.x",
-        "jest": "29.x.x",
+        "jest": "^29.7.0",
         "jest-when": "3.x.x",
         "nodemon": "^3.1.1",
         "ts-jest": "29.x.x",
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -49,11 +49,11 @@ export class Argv {
             writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --shell-isolation does not work with --no-shell-executor-no-image\n`);
         }
 
-        if (argv.defaultImage && argv.shellIsolation) {
+        if (argv.defaultImageExplicitlySet && argv.shellIsolation) {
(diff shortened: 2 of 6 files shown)


src/argv.ts
```diff
@@ -46,8 +46,17 @@ export class Argv {
         argv.injectDotenv(`${argv.cwd}/.gitlab-ci-local-env`, args);
 
         if (!argv.shellExecutorNoImage && argv.shellIsolation) {
-            writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --shell-isolation does not work with --no-shell-executor-no-image\n`);
+            writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --shell-isolation does not work with --shell-executor-no-image\n`);
```
ANGkeith, line 49: not really a typo, 

https://yargs.js.org/docs/#api-reference-booleankey
`--no-shell-executor-no-image` is an alias for `--shell-executor-no-image=false`

src/argv.ts
```diff
@@ -273,6 +282,10 @@ export class Argv {
         return this.map.get("shellExecutorNoImage") ?? true;
     }
 
+    get defaultImage (): string | null {
+        return this.map.get("defaultImage") ?? null;
```
ANGkeith, line 286: image
lets move them here instead

ANGkeith:
```
// https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#container-images
return this.map.get("defaultImage") ?? "docker.io/ruby:3.1";
```
yup, this is what i meant which is mostly how we've set the default args for all the arguments.

ANGkeith (approved):

firecow (approved):

package.json
firecow, line 1: What are these weird changes?

firecow: I'm talking about the specific noop changes to package.json and package-lock.json :smile:

package-lock.json
```diff
@@ -53,7 +53,7 @@
         "depcheck": "1.x.x",
         "esbuild": "^0.24.0",
         "eslint": "8.x.x",
-        "jest": "29.x.x",
+        "jest": "^29.7.0",
```
firecow, line 56:
```suggestion
        "jest": "29.x.x",
```

Issue #1011, opened 2023-10-26
include does not respect variable in rules


netaskd:
**Minimal .gitlab-ci.yml illustrating the issue**
set group/project variable

``` 
# ~/.gitlab-ci-local/variables.yml
group:
  gitlab..example.com/group/:    
    ENABLED_REWRITE: false
# or/and
project:
  gitlab.example.com/group/project.git:    
    ENABLED_REWRITE: false
```
set scripts/.job.yml
```
.job:
  script: echo job
```

set scripts/.rewrite_job.yml
```
.job:
  script: echo rewrite_job
```

set .gitlab-ci.yml
```yml
---
include:
  - local: '/scripts/.job.yml'
  - local: '/scripts/.rewrite_job.yml'
    rules:
      - if: $ENABLED_REWRITE == 'true'

some_job:
  extends:
    - .job

```
instead of content from .job.yml we got content from .rewrite_job.yml

**Expected behavior**
content from .job.yml and echo output "job"

**Host information**
Ubuntu
gitlab-ci-local 4.45.0

**Containerd binary**
docker

**Additional context**
doc is here https://docs.gitlab.com/ee/ci/yaml/includes.html#use-variables-with-include

firecow:
Can't find a dublicate, marking this as a missing feature

ANGkeith:
am i missing out something ? 

i can't replicate the issue

![image](https://github.com/user-attachments/assets/e06c0d99-e5d1-47d8-8548-9ae32168add6)

ANGkeith:
To replicate the issue you'll need 2 projects:

## Project: `gitlab-ci-local-test-2`

**.gitlab-ci.yml**
```yaml
include:
  - local: one.gitlab-ci.yml
    rules:
      - if: '$INCLUDE_LOCAL == "one"'
  - local: two.gitlab-ci.yml
    rules:
      - if: '$INCLUDE_LOCAL == "two"'
```

**one.gitlab-ci.yml**
```yaml
one:
  script:
    - echo "One"
```

**two.gitlab-ci.yml**
```yaml
two:
  script:
    - echo "Two"
```

## Project: `gitlab-ci-local-test`

**.gitlab-ci.yml**
```yaml
include:
  - project: '{user/group}/gitlab-ci-local-test-2'
    ref: main
    file: .gitlab-ci.yml
```

## Variables

[...]

PR #1406, opened 2024-10-30
Respect rules for local includes

ke-lit:
## Explanation

Currently, `rules` are not respected for local includes (#1011).

This PR adds a check for any `rules` in local includes and doesn't include the file if the rules resolve to `never`.

### DIFF
--- a/src/parser-includes.ts
+++ b/src/parser-includes.ts
@@ -94,6 +94,12 @@ export class ParserIncludes {
                     fileDoc["include"] = this.expandInclude(fileDoc["include"], opts.variables);
                     fileDoc["include"].forEach((inner: any, i: number) => {
                         if (!inner["local"]) return;
+                        if (inner["rules"]) {
+                            const rulesResult = Utils.getRulesResult({cwd: opts.cwd, variables: opts.variables, rules: inner["rules"]}, gitData);
+                            if (rulesResult.when === "never") {
+                                return;
+                            }
+                        }
                         fileDoc["include"][i] = {
                             project: value["project"],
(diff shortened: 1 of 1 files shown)


firecow (approved):

ANGkeith:
@ke-lit  could you add test/ sample of gitlab-ci.yml that is not working as expected before this mr is merged

ke-lit:
This turned out not to be as simple an issue as I originally thought. To replicate the issue you'll need 2 projects:

## Project: `gitlab-ci-local-test-2`

**.gitlab-ci.yml**
```yaml
include:
  - local: one.gitlab-ci.yml
    rules:
      - if: '$INCLUDE_LOCAL == "one"'
  - local: two.gitlab-ci.yml
    rules:
      - if: '$INCLUDE_LOCAL == "two"'
```

**one.gitlab-ci.yml**
```yaml
one:
  script:
    - echo "One"
```

**two.gitlab-ci.yml**
```yaml
two:
  script:
    - echo "Two"
```

## Project: `gitlab-ci-local-test`

**.gitlab-ci.yml**
```yaml
include:
  - project: '{user/group}/gitlab-ci-local-test-2'
    ref: main
    file: .gitlab-ci.yml
```

## Variables

Even with `INCLUDE_LOCAL` not set in `~/.gitlab-ci-local/variables.yaml`, both jobs `one` and `two` run when they shouldn't. Setting `INCLUDE_LOCAL` to `one`, `two` or any other value has no impact.

ANGkeith:
cool! got it, thanks, will add the test case when i'm free

PR #1421, opened 2024-11-11
Fix list-json rules by only returning rules if defined

ekohilas:
list-json was returning `null` in json for rules output when the rules were undefined


ANGkeith:
i believe current behavior should be correct 

ie.
```yaml
job:
  script:
    - echo "Hello world!"

---
job-2:
  rules: [] # This job will never be executed
  script:
    - echo "Hello world!"
```

ekohilas:
Ah that's a good perspective. Thanks I'll update for that :blush:

ANGkeith (approved):

PR #1422, opened 2024-11-11
Fix running error by reverting import with type json

ekohilas:
When running the tool, the following error is returned:
```
npx --yes gitlab-ci-local@latest                                                                                                                                                 
file:///Users/username/.npm/_npx/3f18f2096def1b34/node_modules/gitlab-ci-local/src/index.js:13
import packageJson from "../package.json" with { type: "json" };
                                          ^^^^

SyntaxError: Unexpected token 'with'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)
    at async link (node:internal/modules/esm/module_job:68:21)

Node.js v18.18.2
```

**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
job:
  script:
    - echo "Heya"
```

**Expected behavior**
Tool to run

**Host information**
MacOs
gitlab-ci-local 4.55

**Containerd binary**
Are you using docker

**Additional context**
This seems related to #1331 and #1337

My installed typescript version is 5.6.3, so I don't think it's that.

Still trying to figure out why this doesn't work.

PR #1424, opened 2024-11-11
Fix json import

ekohilas:
Reopening of #1422 

In Node 18.18, the following error occurs:
```
npx --yes gitlab-ci-local@latest                                                                                                                                                 
file:///Users/username/.npm/_npx/3f18f2096def1b34/node_modules/gitlab-ci-local/src/index.js:13
import packageJson from "../package.json" with { type: "json" };
                                          ^^^^

SyntaxError: Unexpected token 'with'
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:119:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:468:14)
    at async link (node:internal/modules/esm/module_job:68:21)

Node.js v18.18.2
```

This PR fixes the issue by converting json imports to instead use readfilesync and JSON.parse instead, which makes it more explicit and avoids this error.

This fix allows users who may otherwise not be able to update to Node v18.20 overnight to continue using this tool.

Thanks again!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 import chalk from "chalk";
+import {readFileSync} from "fs";
 import yargs from "yargs";
 import {Parser} from "./parser.js";
 import * as state from "./state.js";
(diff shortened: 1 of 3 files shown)


ANGkeith (approved):

firecow:
@ANGkeith You sure we wanna keep node 18 support?

ANGkeith:
hmm not really, but at least for now, does'nt seem like there is too much incompatibility.

PR #1439, opened 2024-11-26
Use re2js, its faster and is much less complex build-wise

firecow:


firecow:
@ANGkeith I managed to get it working, yay 😄

src/utils.ts
```diff
@@ -209,6 +210,20 @@ export class Utils {
         if (ruleIf === undefined) return true;
         let evalStr = ruleIf;
 
+        const flagsToBinary = (flags: string): number => {
+            let binary = 0;
+            if (flags.includes("i")) {
+                binary |= RE2JS.CASE_INSENSITIVE;
+            }
+            if (flags.includes("g")) {
```
ANGkeith, line 218:
```suggestion
            if (flags.includes("s")) {
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll

src/global.ts
```diff
@@ -0,0 +1,18 @@
+import {RE2JS} from "re2js";
+
+declare global {
+    interface String {
+        matchRE2JS(o: RE2JS): Array<string> | null;
+    }
+}
+String.prototype.matchRE2JS = function (o: RE2JS): Array<string> | null {
```
ANGkeith, line 8: hacky, but coool

src/global.ts
```diff
@@ -0,0 +1,18 @@
...
+    let results: string[] | null = null;
+    const matcher = o.matcher(this.toString());
+    while (matcher.find()) {
+        const g = matcher.group();
+        if (g == "") continue;
+        results = results == null ? [] : results;
```
ANGkeith, line 14:
```suggestion
        results = results ?? [];
```

as recommended by sonar

ANGkeith (approved):

ANGkeith:
> @ANGkeith I managed to get it working, yay 😄

very nice! lgtm

ANGkeith:
tho, just note that https://github.com/le0pard/re2js?tab=readme-ov-file#performance, the node-re2 be more performant.

so the main benefit is the simplified build process

firecow:
> tho, just note that https://github.com/le0pard/re2js?tab=readme-ov-file#performance, the node-re2 be more performant.
> 
> so the main benefit is the simplified build process

Runtime performance isn't the issue with using https://github.com/uhop/node-re2/, it's the startup time.

PR #1444, opened 2024-11-30
chore: rename integration test file name

ANGkeith:
small clean up, we don't really need a unique name for our test file 

this will prevent typo like this
image


firecow (approved):
:heavy_check_mark:

PR #1460, opened 2024-12-23
feat: get interactive shell into jobs running in a container

### LINKED ISSUE #1032: Allow entering the container to debug its state
I am trying to debug a crash in a process using this tool. I'd need to run again one of the processes with gdb and interact with it.

I'd like to be offered a way to interact with the system as one of the tests fails.

I looked into modifying our ci tooling to call the processes by default with gdb, can't really do that without breaking our current ci as we get the recipes from elsewhere shared.

**Additional context**
This is the CI in question:
https://invent.kde.org/plasma/kwin/-/blob/Plasma/5.27/.gitlab-ci.yml?ref_type=heads

KoiFresh:
This MR adds the ability to use the `# @Interactive` decorator also on jobs containing an `image` tag. Also the job does no longer have to open the shell by itself, instead you'll get a shell instead of executing the script. In the shell the script can be run with `/gcl-cmd`.

Its also possible to run a job in interactive mode with `-i <job name>` or `--interactive-jobs <job name>` to get a shell into a job without modifying the `.gitlab-ci.yml`, or run with `--debug` to get a shell into jobs which failed. 

Fixes the glob pattern, because the syntax for the local include pattern used by [gitlab](https://docs.gitlab.com/ee/ci/yaml/includes.html#use-includelocal-with-wildcard-file-paths) defers from glob pattern used by [globby](https://github.com/sindresorhus/globby).

Closes #1032


fschrempf:
I just had a try with this and it works just fine for me. I tested the decorator and `--debug`.

firecow:
@KoiFresh Feel free to reopen, if you find time to complete this task

Issue #1447, opened 2024-12-03
Add a new option --force-shell-executor which would always ignore images and always run the shell executor


ecourreges-orange:
**Is your feature request related to a problem? Please describe.**
I would like to run gitlab-ci-local from a devcontainer which already has everything needed to build my project and has the same dependencies as the different partial images used in different pipeline jobs.  
To avoid adding a level of complexity with the docker executor which is used by default when there is either a job image or a default image, **I would like to add the possibility to force the shell executor for the projects that I choose.**

Similar topic from #1064 and #1314 but easier to implement.

**Describe the solution you'd like**
**I would add an option --force-shell-executor** to the command line which basically always sets imageName to `null`, I think it should be easy enough to code and should work, and I also hope it works with env conf so that I set it once and for all for each project where I need the option.

I am willing to provide the PR after discussion, once you are ok with the concept.

**Describe alternatives you've considered**
Loosing execution time and complexity by running the default docker executor

**Additional context**
From a devcontainer that contains a C++/Bazel or a Java/Maven build system, I want to be able to compile, docker build, deploy locally on minikube in iterations of less than 30s, so each second lost launching a docker is not acceptable.

Thanks for your time,
Regards.

ecourreges-orange:
It would almost just need to be this, adding   `if (this.argv.shellExecutor) return null;` to the imageName function:
```typescript
    private imageName (vars: {[key: string]: string} = {}): string | null {
         if (this.argv.shellExecutor) {
            return null;
        }
        const image = this.jobData["image"];
        if (!image) {
            if (this.argv.shellExecutorNoImage) {
                return null;
            } else {
                // https://docs.gitlab.com/ee/ci/runners/hosted_runners/linux.html#container-images
                return this.argv.defaultImage;
            }
        }
        const expanded = Utils.expandVariables(vars);
        const imageName = Utils.expandText(image.name, expanded);
        return imageName.includes(":") ? imageName : `${imageName}:latest`;
    }
```

PR #1481, opened 2025-01-27
feat: add --shell-executor option. (#1447).

ecourreges-orange:
--shell-executor option (#1447)
Allows forcing to always use the shell executor even when there are images declared somewhere. Avoids image downloads and updates when you want a speedy pipeline

### DIFF
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -58,6 +58,10 @@ export class Argv {
             writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --default-image does not work with --shell-executor-no-image=true\n`);
         }
 
+        if (argv.defaultImageExplicitlySet && argv.forceShellExecutor) {
+            writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --default-image does not work with --force-shell-executor=true\n`);
+        }
+
         return argv;
     }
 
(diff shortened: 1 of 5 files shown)


src/argv.ts
```diff
@@ -50,6 +50,10 @@ export class Argv {
             writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --shell-isolation does not work with --no-shell-executor-no-image\n`);
         }
 
+        if (!argv.shellExecutor && argv.shellIsolation) {
+            writeStreams?.stderr(chalk`{black.bgYellowBright  WARN } --shell-isolation does not work with --no-shell-executor\n`);
```
firecow, line 54: I run `gitlab-ci-local --shell-isolation` all the time, wouldn't this break my workflow?
Given that `argv.shellExecutor` defaults to false.

ecourreges-orange: Indeed, didn't think it through, removing this and the associated test

src/index.ts
```diff
@@ -172,6 +172,11 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
             description: "Enable artifact isolation for shell-executor jobs",
             requiresArg: false,
         })
+        .option("shell-executor", {
```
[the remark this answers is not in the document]
ecourreges-orange: Alright I renamed everything to force-shell-executor which has the benefit of being explicit in all cases.

ANGkeith (approved):

PR #1487, opened 2025-01-30
fix: (#1486) remove useless remoteFileExists check at every launch

### LINKED ISSUE #1486: Speedup component download by removing useless remote file check on startup
```yml
include:
  # Terraform template
  - component: "$CI_SERVER_FQDN/dds/Common/ci/terraform/gitlab-ci-terraform@feat/multiAZ"
  # Terraform template (Google Cloud variant)
  - component: "$CI_SERVER_FQDN/BAD_URL/BAD_COMPONENT@bad_version"
  # WIF template - Orange configuration
  - project: 'multicloud-patterns/tools/gcp-gitlab-template'
    ref: 'main'
    file: '/templates/gitlab-ci-wif-orange-configuration.yml'
  # gitlab-ci-local wif-token
  - component: "$CI_SERVER_FQDN/dds/common/wif-token/auto-refresh@1.0"
    inputs:
      auto-refresh-enabled: "true"
```

**Expected behavior**
Slow first download (~10s), fast subsequent launches (<3s)

**Real Behaviour**
Takes at least 7s every time.

The culprit is a remote file check **at every launch** which breaks the initial promise of "your components are not downloaded again once they are in .gitlab-ci-local"

**Proposed Fix**
I will post a PR that removes an unwanted check that the remote file of the component exists, as removing that still provides a legible error in case it doesn't exist, as the real download happens later in the code.
I will be removing [lines 150-152 of parser-includes.ts](https://github.com/firecow/gitlab-ci-local/blob/master/src/parser-includes.ts#L150)

Maybe also the function `remoteFileExist` seems to become dead code but I will not remove it on first proposal.

Error looks like this in existing code which takes here 6s:
```
 WARN  The job: `trigger_downstream_pipeline` will be no-op. Multi-project pipeline is not supported by gitlab-ci-local
Error: Error: Command failed with exit code 128: git archive --remote=ssh://git@gitlab.tech.orange:22/BAD_URL.git bad_version templates/BAD_COMPONENT.yml
remote:
remote: ========================================================================
remote:
remote: The namespace you were looking for could not be found.
remote:
remote: ========================================================================
remote:
fatal: the remote end hung up unexpectedly
    at Function.remoteFileExist (/snapshot/firecow-gitlab-ci-local/src/utils.ts:350:112)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Function.init (/snapshot/firecow-gitlab-ci-local/src/parser-includes.ts:150:31)
    at async e.init (/snapshot/firecow-gitlab-ci-local/src/parser.ts:109:44)
    at async Function.create (/snapshot/firecow-gitlab-ci-local/src/parser.ts:60:9)
    at async _c (/snapshot/firecow-gitlab-ci-local/src/handler.ts:55:18)
    at async Object.handler (/snapshot/firecow-gitlab-ci-local/src/index.ts:33:21)
gitlab-ci-local --list  0.90s user 0.20s system 17% cpu 6.192 total
```

Error looks like this with the code removed which takes 4s but in success cases less:
```
╰─ time gcil --list
 WARN  The job: `trigger_downstream_pipeline` will be no-op. Multi-project pipeline is not supported by gitlab-ci-local
Project include could not be fetched { project: BAD_URL, ref: bad_version, file: templates/BAD_COMPONENT.yml }
Error: Command failed with exit code 2: set -eou pipefail; git archive --remote=ssh://git@gitlab.tech.orange:22/BAD_URL.git bad_version templates/BAD_COMPONENT.yml | tar -f - -xC .gitlab-ci-local/includes/gitlab.tech.orange/BAD_URL/bad_version/
remote:
remote: ========================================================================
remote:
remote: The namespace you were looking for could not be found.
remote:
remote: ========================================================================
remote:
fatal: the remote end hung up unexpectedly
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
gitlab-ci-local --force-shell-executor --list  0.91s user 0.16s system 22% cpu 4.685 total
```

ecourreges-orange:
This remoteFileExists check at every launch was slowing down all launches which include components.
See Issue (#1486) for more details.

fixes #1486

### DIFF
--- a/src/parser-includes.ts
+++ b/src/parser-includes.ts
@@ -129,14 +129,25 @@ export class ParserIncludes {
                 }
             } else if (value["component"]) {
                 const {domain, port, projectPath, componentName, ref} = this.parseIncludeComponent(value["component"]);
-                // converts component to project
-                const files = [`${componentName}.yml`, `${componentName}/template.yml`, null];
+                // converts component to project. gitlab allows two different file path ways to include a component
+                let files = [`${componentName}.yml`, `${componentName}/template.yml`, null];
+                const isLocalComponent = projectPath === `${gitData.remote.group}/${gitData.remote.project}` && ref === gitData.commit.SHA;
+
+                // If a file is present locally, keep only that one in the files array to avoid downloading the other one that never exists
+                if (!argv.fetchIncludes) {
+                    for (const f of files) {
+                        const localFileName = `${cwd}/${stateDir}/includes/${gitData.remote.host}/${projectPath}/${ref}/${f}`;
+                        if (fs.existsSync(localFileName)) {
+                            files = [f];
+                            break;
+                        }
+                    }
+                }
 
                 for (const f of files) {
                     assert(f !== null, `This GitLab CI configuration is invalid: component: \`${value["component"]}\`. One of the files [${files}] must exist in \`${domain}` +
                                         (port ? `:${port}` : "") + `/${projectPath}\``);
 
-                    const isLocalComponent = projectPath === `${gitData.remote.group}/${gitData.remote.project}` && ref === gitData.commit.SHA;
                     if (isLocalComponent) {
                         const localComponentInclude = `${cwd}/${f}`;
                         if (!(await fs.pathExists(localComponentInclude))) {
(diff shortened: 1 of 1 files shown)


ANGkeith (approved):

src/parser-includes.ts
```diff
@@ -129,14 +129,23 @@ export class ParserIncludes {
...
+                const isLocalComponent = projectPath === `${gitData.remote.group}/${gitData.remote.project}` && ref === gitData.commit.SHA;
+
+                // If a file is present locally, keep only that one in the files array to avoid downloading the other one that never exists
+                for (const f of files) {
+                    const localFileName = `${cwd}/${stateDir}/includes/${gitData.remote.host}/${projectPath}/${ref}/${f}`;
+                    if (fs.existsSync(localFileName)) {
+                        files = [f];
+                        break;
+                    }
+                }
```
ANGkeith, line 143: I'm thinking about performing this only when `fetchIncludes == false` (--no-fetch-includes) , what do you think ?

ecourreges-orange: Thanks, good catch, yeah at least for the case where somehow a template is switched from `templates/mycomponent.yml`  to `mycomponent/template.yml` it would help a bit, although in that case, if the `.gitlab-ci-local/` were not cleaned you would end up with the 2 files which is as far as I saw undefined as of Gitlab documentation.  
I will do the change, supposing that fetch-includes pre-cleans each include.

ANGkeith (approved):

Issue #1473, opened 2025-01-21
Compatibilty with running the Debian Salsa CI locally


ottok:
This is halfway between a bug report and a feature request. Since the latest update to https://salsa.debian.org/salsa-ci-team/pipeline#running-the-pipeline-locally this gitlab-ci-local is now advertised as a potential tool for running the Salsa CI pipeline locally. Salsa CI is Debian's custom GitLab CI pipeline for testing Debian packaging git repositories before uploading Debian source packages to Debian archives.

I am however not able to fully run the pipeline locally, as the `RELEASE` variable does not propagate properly.

Steps to reproduce on a Debian/Ubuntu host:

```
# Install git-buildpackage
sudo apt install git-buildpackage --no-install-recommends
# Clone Debian packaging sources of an example package
gbp clone --verbose --add-upstream-vcs https://salsa.debian.org/debian/entr.git
cd entr
# Preview seems to work correctly
gitlab-ci-local --file debian/salsa-ci.yml --preview
# The list seems mostly correct, but some test jobs are missing
gitlab-ci-local --file debian/salsa-ci.yml --list
name                              description  stage                       when        allow_failure  needs
extract-source                                 provisioning                on_success  false      
build                                          build                       on_success  false      
build armel                                    build                       on_success  false      
build armhf                                    build                       on_success  false      
build arm64                                    build                       on_success  false      
build riscv64                                  build                       on_success  false      
build source                                   build                       on_success  false      
upgrade from bookworm                          upgrade package and distro  on_success  false      
upgrade from bullseye                          upgrade package and distro  on_success  false      
upgrade from buster                            upgrade package and distro  on_success  false      
upgrade from stretch                           upgrade package and distro  on_success  false
# Run the entire pipeline locally
gitlab-ci-local --file debian/salsa-ci.yml
```

The first job `extract-sources` passes correctly, but the second job `build` fails on:
```
parsing and downloads finished in 318 ms.
json schema validated in 133 ms
build starting registry.salsa.debian.org/salsa-ci-team/pipeline/base: (build)
Error: Command failed with exit code 125: docker pull registry.salsa.debian.org/salsa-ci-team/pipeline/base:
Error: invalid reference format
    at makeError (/snapshot/firecow-gitlab-ci-local/node_modules/execa/lib/error.js:60:11)
    at <anonymous> (/snapshot/firecow-gitlab-ci-local/node_modules/execa/index.js:118:26)
[...]

PR #1489, opened 2025-02-02
feat: define CI_BUILDS_DIR predefined variable

ANGkeith:


ottok:
Should the git commit message explain *why* you suggest this change?

I have seen issues with the path but this commit does not convince why defining it like this is the best solution.

ANGkeith:
> Should the git commit message explain why you suggest this change?

We're updating the predefined variable on a best-effort basis, mostly mirroring GitLab's implementation ([GitLab Docs](https://docs.gitlab.com/ci/variables/predefined_variables/)). Since we're just following their updates, there may not be much to explain beyond that.


> but this commit does not convince why defining it like this is the best solution.
yeah, i don't think this is the best solution, we probably need to update 
https://github.com/firecow/gitlab-ci-local/blob/master/src/job.ts#L22  accordingly and update CI_BUILDS_DIR -> /builds

ottok:
In #1473 I noticed that one needs to pass `--variable CI_BUILDS_DIR=/gcl-builds` to the command. This PR is likely to fix that, not matter what path you end up choosing, as having that variable defined will make the paths work.

firecow (approved):

PR #1532, opened 2025-03-19
style: upgrade eslint and enforce consistent linebreak style for operators

ANGkeith:
Not sure if there's something wrong with my setup.. 

Locally `npm run eslint` i am always getting the following error
```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=4.7.4 <5.6.0

YOUR TYPESCRIPT VERSION: 5.8.2

Please only submit bug reports when using the officially supported version.

=============
```

Instead, i've decided to upgrade eslint to v9

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,0 +1,51 @@
+import eslint from "@eslint/js";
+import tseslint from "typescript-eslint";
+import stylisticTs from "@stylistic/eslint-plugin-ts";
+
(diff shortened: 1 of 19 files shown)


firecow (approved):
Very nice, thanks mate :smile:

PR #1555, opened 2025-04-23
fix: Fix inputs parsing when ctx.inputs is undefined.

batk0:
A tiny fix for parsing inputs. It's for the case when `spec.inputs` are specified in main `.gitlab-ci.yml` with defaults and no inputs provided for the job.

Example:
```
---
spec:
  inputs:
    foo:
      type: string
      default: bar
---
job:
  script:
    - echo $[[ inputs.foo ]]
```


ANGkeith (approved):

ANGkeith:
@batk0  could you fix the failing test

other than that lgtm

batk0:
@ANGkeith Done

PR #1560, opened 2025-04-30
fix: Address wildcard path resolution in local include for project fi…

hoolos:
Address wildcard path resolution in local include for project file ref includes


firecow:
@hoolos Feel free to reopen if you find time to complete this PR

PR #1566, opened 2025-05-14
Implement support for podmans userns option

dwagenk:
I had problems when running a CI pipeline via gitlab-ci-local on a host that uses podman as container engine. Experimenting with the podman invocation locally (the container init script does some  I found out, that using the `--userns=keep-id` [docs](https://docs.podman.io/en/v4.6.1/markdown/options/userns.container.html) option fixes the problem.

So this adds the flag needed to let gitlab-ci-local with the needed flag in the container executor invocation.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -202,6 +202,11 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
             description: "Sets docker user to 0:0",
             requiresArg: false,
         })
+        .option("userns", {
+            type: "string",
+            description: "Set docker executor userns option",
+            requiresArg: false,
+        })
         .option("privileged", {
             type: "boolean",
             description: "Set docker executor to privileged mode",
--- a/src/argv.ts
+++ b/src/argv.ts
@@ -207,6 +207,10 @@ export class Argv {
         return this.map.get("umask") ?? true;
     }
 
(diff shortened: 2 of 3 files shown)


src/index.ts
```diff
@@ -202,6 +202,11 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
             description: "Sets docker user to 0:0",
             requiresArg: false,
         })
+        .option("userns", {
+            type: "string",
+            description: "Passes the argument to podmans --userns option",
```
firecow, line 207:
```suggestion
            description: "Set docker executor userns option",
```

dwagenk: Would be OK for me, but AFAIK docker does not have the `userns` option, only podman does. I thought about adding a check for this option so it can only be combined with podman es executor, but did not see any other parameter being handled that way.

firecow:
```
mjn@mjn-lapdows:~$ docker run --help | grep userns
      --userns string                    User namespace to use
```

dwagenk: Added your suggestion and addressed the [eslint finding](https://github.com/firecow/gitlab-ci-local/actions/runs/15021932826/job/42263206098).

firecow (approved):

PR #1573, opened 2025-05-20
Escape variables in image entrypoints

### LINKED ISSUE #1572: Variable substitution should not happen in the job's entrypoint
```yml
image-entrypoint-with-variables:
  variables:
    FOO: BAR
  image:
    name: alpine
    entrypoint:
      - sh
      - -c
      - |
        if [ -z "$FOO" ]; then >&2 echo FOO is empty in entrypoint; exit 1; fi
        sh "$@"
  script:
    - echo "success"
```

**Expected behavior**
`"$FOO"` resolves to `"BAR"` when the entrypoint is executed and the script runs
https://gitlab.com/jonathang/test/-/jobs/10097228369
```
$ echo "success"
success
```

**Current behavior**
```
image-entrypoint-with-variables starting alpine:latest (test)
image-entrypoint-with-variables copied to docker volumes in 623 ms
image-entrypoint-with-variables > FOO is empty in entrypoint
image-entrypoint-with-variables finished in 1.28 s  FAIL 1 
```

The `$FOO` variable is substituted with the value from the environment of the calling process.
We can see it from the container command: 
```json
"Cmd": [
  "-c",
  "if [ -z  ]; then >&2 echo FOO is empty in entrypoint; exit 1; fi\nsh \n",
  ...
``` 

**Host information**
nixos
gitlab-ci-local 4.59.0 and latest commit from master branch

**Containerd binary**
docker

Meallia:
fixes #1572 

During the creation of the docker command to execute the job, the entrypoint arguments were passed to bash surrounded by double quotes.

Bash would then substitute variables (like `${FOO}` or `$BAR`), commands (like `$(cmd)` and `` `cmd` ``) in the context of the gitlab-ci-local process.

By surrounding those arguments with single quotes and escaping them in the strings, bash won't try to do anything fancy with the strings and leave them as-is.


Meallia:
while doing some more testing, I realized all entrypoint arguments are not properly escaped. putting this back to draft for now.

ANGkeith (approved):

ANGkeith:
nice, lgtm

firecow:
@Meallia is this ready?

Meallia:
Yes, it's ready.

PR #1657, opened 2025-10-07
feat: Add validate dependency chain flag to validate broken dependencies without running the pipeline

tushar-gupta-1995:
### Overview: 
Adds a new --validate-dependency-chain flag that validates job dependencies in GitLab CI pipelines to catch configuration errors before pipeline execution. Excludes external project dependencies, pipeline dependencies.

###   What are dependency chains in this context?
Dependency chains refer to the relationships between GitLab CI jobs defined using the either needs or dependencies keyword, where one job depends on another job to complete before it can start OR it needs artifact from a job in previous stage. For example:

```
build-job:
    script: echo "Building..."

  test-job:
    needs: [build-job]  
    script: echo "Testing..."

  deploy-job:
    needs: [build-job, test-job] 
    script: echo "Deploying..."
```

This creates a dependency chain: build-job → test-job → deploy-job

  The validation catches scenarios like (without actually executing the pipeline):
  - deploy-job needs missing-job (job doesn't exist)
  - test-job needs build-job but build-job has when: never
  - deploy-job needs build-job but rules exclude build-job from running
  
###   What does this MR do?

  - Introduces --validate-dependency-chain CLI flag
  - Validates that all jobs referenced in needs declarations exist and will be active in the current pipeline execution, without running them.
  - Provides clear error messages showing which jobs have missing dependencies
  - Exits with descriptive error when validation fails/

###   Why we are introducing this change

  - The linter offered by gitlab only validates dependency chains on push to main but not on other events triggered based on predefined and user defined gitlab variables
  - Its easy for a reviewer and developer to miss failure scenarios described above specially on custom events

### Example of failure detection

```
tushar.gupta@mac-1010-tg test-result-tracking % gitlab-ci-local --list --variable CI_PIPELINE_SOURCE=schedule --variable CI_COMMIT_BRANCH=main --variable CI_DEFAULT_BRANCH=main  --validate-dependency-chain 
 WARN  Avoid overriding predefined variables (​https://gitlab.com/gitlab-org/gitlab/-/blob/v17.7.1-ee/doc/ci/variables/predefined_variables.md?plain=1&ref_type=tags#L15-16​) [CI_PIPELINE_SOURCE,CI_COMMIT_BRANCH,CI_DEFAULT_BRANCH] as it can cause the pipeline to behave unexpectedly.
parsing and downloads finished in 255 ms.
json schema validated in 109 ms
name                            description  stage                   when        allow_failure  needs
assign_projects                              test                    always      false      
bump version                                 version                 on_success  false          [prepare artifact version]
  bump version needs prepare artifact version which does not exist
Dependency chain validation will fail with event: CI_PIPELINE_SOURCE=schedule CI_COMMIT_BRANCH=main CI_DEFAULT_BRANCH=main
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -97,7 +97,7 @@ process.on("SIGUSR2", async () => await cleanupJobResources(jobs));
         })
         .option("validate-dependency-chain", {
             type: "boolean",
-            description: "Validate that jobs needed or dependent by active jobs under specified conditions are also active without actually running the jobs. If validation fails, use --list flag to see which jobs will run under specified conditions",
+            description: "Validate that jobs needed or dependent by active jobs under specified conditions are also active without actually running the jobs. Uses fail-fast approach - stops at first validation error for both 'needs' and 'dependencies' keywords. If validation fails, use --list flag to see which jobs will run under specified conditions",
             requiresArg: false,
         })
         .option("preview", {


src/commander.ts
```diff
@@ -272,4 +272,47 @@ export class Commander {
         });
     }
 
+    static validateDependencyChain (parser: Parser, writeStreams: WriteStreams): boolean {
```
firecow, line 275: Isn't this a complete dublicate of the code in https://github.com/firecow/gitlab-ci-local/blob/master/src/executor.ts

PR #1671, opened 2025-10-28
feat(yargs): update yargs to 18.0.0

oldium:
New `yargs` version bundling is failing, because it is newly ESM-only module:

 * It is not included in `esbuild` and `pkg` does not patch [ESM import paths][1].
 * If added into `esbuild`, the transpiled `import.meta.url` is empty and execution fails.

Fix this by letting `esbuild` transpile also `yargs` to CJS, but provide a sensible value for `import.meta.url` as mentioned in the `esbuild` [bug report][2].

While at it, also update `@yao-pkg/pkg` to support newer Node.js runtime.

[1]: https://github.com/yao-pkg/pkg/issues/16#issuecomment-2334512347
[2]: https://github.com/evanw/esbuild/issues/2441#issuecomment-1315341698

Supercedes: #1577


oldium:
Ha! The `yargs` locales are probably missing due to this.

oldium:
Hm. I am able to get also locales working locally, but when packed with `pkg`, the `--help` option gets to segmentation fault... I have no solution for this.

oldium:
I will leave it here in this state, feel free to close it if you do not want to continue on it.

oldium:
Maybe try Node.js experimental Single Executable Applications?

firecow:
We've started using bun. So yargs 18 is coming soon

PR #1695, opened 2025-11-26
feat: add the ability to define a custom ca certificate for docker images

harryvince:
Adds the ability to define a custom ca cert for using the tooling, useful for when things such as MITM certs are in place.

- Use via CLI: `--ca-file <path>`
- Env Variable: `GCL_CA_FILE`
- .env support: `CA_FILE=<path>`


harryvince:
@firecow have resolved the yamllint failure so should be good to trigger the workflow again 🙂

Issue #1700, opened 2025-11-29
Utils.remoteFileExist broken for http(s) on instances with login requirements


kk-chesse:
Follow up from #1698.  
`Utils.remoteFileExist` uses axios to check if a file exists on http(s) repositories. Now with  #1699 it always fails, due to being redirected to the signin page. 

**Additional context**
Gitlab 18.6 configured to require login on http(s)

I'm not sure if this is a use case you want to support. If not, at least the behaviour should be documented.

I looked into potential fixes involving using the git client to avoid login problem.
1. Use git archive with https, this requires atleast git 2.44 (released February 2024). Unfortunately this didn't work for me
  `error: RPC failed; HTTP 422 curl 22 The requested URL returned error: 422`
2. Do a sparse checkout git 2.25 (released January 2020). While a lot more complicated, it could unify the logic for all protocols supported by git

kk-chesse:
Looking into the usage of `Utils.remoteFileExist` I see it's only used once in `ParserIncludes.init` to determine the directory structure of the component repository.

Thinking out loud, would it be feasible to refactor the `ParserIncludes.init` to do a sparse checkout with both `templates/<component>.yml` and `templates/<component>/template.yml` and check later which file to use? This would reduce the required calls to the server to just 1, instead of 2 or 3.

ANGkeith:
can you show the output of your redacted `git remote -v` ? 

i think you're currently using `https` for your repo right ? 

could you update it to `ssh` instead ?

ie.
```
# Don't use https
git remote -v
origin  https://gitlab.com/example/apple.git (fetch)
origin  https://gitlab.com/example/apple.git (push)

# use this instead
git remote -v
git@gitlab.com:ANGkeith/shhh.git
origin  git@gitlab.com:example/apple.git (fetch)
origin  git@gitlab.com:example/apple.git (push)
```

Then your issue should be fixed.


---
yeah, i totally forgotten. The remoteFileExist will not work for repos that are private

kk-chesse:
```
git remote -v
origin  https://<domain>/group/project.git (fetch)
origin  https://<domain>/group/project.git (push)
```

I can't use ssh. Any operations just sits idle for a few minutes and then timeouts. I think because our coporate firewall blocks it. 

I've have a proof of concept to use sparse checkouts for components #1701

peterbbeu:
[...]

Issue #1319, opened 2024-08-22
Crash with gitlab-ci component at specific commit


pixelshot91:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
include:
  - component: <my_gitlab_component_url>@<specific_commit_hash>
```

```shell
$ gitlab-ci-local
Error: Error: Command failed with exit code 1: git archive --remote=ssh://git@<my_gitlab_component_git_url> <specific_commit_hash> templates/common.yml
fatal: sent error to the client: git upload-archive: archiver died with error
remote: fatal: no such ref: <specific_commit_hash>        
remote: git upload-archive: archiver died with error
    at Function.remoteFileExist (/snapshot/firecow-gitlab-ci-local/src/utils.ts:289:112)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Function.init (/snapshot/firecow-gitlab-ci-local/src/parser-includes.ts:126:31)
    at Parser.init (/snapshot/firecow-gitlab-ci-local/src/parser.ts:109:44)
    at Function.create (/snapshot/firecow-gitlab-ci-local/src/parser.ts:60:9)
    at handler (/snapshot/firecow-gitlab-ci-local/src/handler.ts:78:18)
    at Object.handler (/snapshot/firecow-gitlab-ci-local/src/index.ts:37:21)
```

The problem is caused by `git archive`. I get the same error if I launch it manually
```shell
$ git archive --remote=ssh://git@<my_gitlab_component_git_url>  <specific_commit_hash> templates/common.yml
fatal: sent error to the client: git upload-archive: archiver died with error
remote: fatal: no such ref: <specific_commit_hash>
remote: git upload-archive: archiver died with error
```

If I use a branch name or no ref at all, the CI works correctly. But a specific commit hash is the only way to guarantee reproducibility.

It seems that I is intentionally forbidden by git to get a file at a specific commit.

Should we use `git clone && git checkout` to avoid this problem ?

**Expected behavior**
The component should be correctly included

**Host information**
Debian 12.6
gitlab-ci-local 4.52.2

ANGkeith:
>  But a specific commit hash is the only way to guarantee reproducibility.

regarding this point, as an workaround to our implementation limitation, you can use `git tag` which should work with `git archive --remote`

ANGkeith:
I won't be implementing this since we should be able to live with the workaround. 

But for future note, do take note of the following potential performance issue when dealing with large repos

[...]

PR #1701, opened 2025-11-29
Use sparse checkout for components

kk-chesse:
This PR reworks how components are pulled from the remote. Which would fail, if you tried to pull files from a private project with http(s)

**Old way**
1. check if file exists on remote, to determine which directory structure to use (this steps would fail on private repositories)
2. "convert" component to project include 
3. download as project include
4. handle as project include

**New way**
1. download components with a sparse checkout, download both possible locations in the same call (this way we avoid login requirements)
2. check if the included component exists
3. handle as component include

**Additional changes**
I had to change how `parseIncludeComponent` fetches to the tags. From try/catch to if/else, because if a firewall blocks ssh it takes a few minutes before the call timeouts...

**TODOs**
- ~~I'm calling `parseIncludeComponent` twice, which is expensive, due to fetching tags~~
  Done, the result is cached now
- ~~I'm not sure if the inner expands done for projects are neccessary for components~~
  Done, code is refactored so it can be used for project and component includes
- ~~This PR remove the only use of `Utils.remoteFileExist`, but I don't know if you want to keep it around~~
  Done, function is removed

### DIFF
--- a/src/parser-includes.ts
+++ b/src/parser-includes.ts
@@ -357,7 +357,7 @@ export class ParserIncludes {
                     "git checkout",
                     `cd ${cwd}/${stateDir}`,
                     `mkdir -p ${tmpDir}/templates`, // create templates subdir (if it doesn't exist), as the check out may not create it
-                    `cp -r ${tmpDir}/templates ${cwd}/${target}/templates`,
+                    `cp -r ${tmpDir}/templates ${cwd}/${target}`,
                 ], cwd);
             } else {
                 // git archive fails if the paths do not exist, to work around this we use a wildcard "templates/component*.yml"


ANGkeith:
using `git` sounds good to me. That way the storing of creds(like PAT) can be handled by `git` 

these refactoring won't be easy, we have a couple of integration test, make sure those passes to prevent regression bugs

kk-chesse:
I worked on the next version to. Changes;

- fixed the integration tests
- fixed the missing expands in inner local includes
- added a cache for the parsed components

The tests are working on my end and I can use it with our private Gitlab instance.

ANGkeith:
nice work man, thanks for the contribution. in general lgtm, but will take another look when i have more spare time.
good to see all the integration test passes without much tweaking

I see you have TODOs in your pr description.

>1.  I'm calling parseIncludeComponent twice, which is expensive, due to fetching tags
>1. I'm not sure if the inner expands done for projects are neccessary for components
>1.  This PR remove the only use of Utils.remoteFileExist, but I don't know if you want to keep it around


re. no.1  i think you have resolved it right? 
re. no.2 what do you mean here? 
re. no.3 i'm ok with removing that `Utils.remoteFileExist`

----
On a seperate topic, 

i also see that you've considered using git archive. 

a while back, i did some [profiling](https://github.com/firecow/gitlab-ci-local/issues/1319#issuecomment-2316587497) and found that git archive can be more performant on elephant size repo like gitlab.com

interestingly, i face different error from you when attempting git archive via https. 
getting the same error same as [this](https://stackoverflow.com/a/79418858) but yeah, guess we have no choice but to go with sparse checkout

kk-chesse:
I did a last round of refactoring and removed `Utils.remoteFileExist`. I think it's ready for review now.

>re. no.1 i think you have resolved it right?  

Yes, now the result is cached
>re. no.2 what do you mean here?

That TODO was mostly for myself. I wasn't sure what some code did, but now I know and refactored the code accordingly.
>re. no.3 i'm ok with removing that Utils.remoteFileExist

Done :)

I will update the opening post

---
Yeah, it's a shame `git archive` isn't working for https.

> a while back, i did some [profiling](https://github.com/firecow/gitlab-ci-local/issues/1319#issuecomment-2316587497) and found that git archive can be more performant on elephant size repo like gitlab.com

Super interesting profiling data, thank you for showing me that.

src/parser-includes.ts
```diff
@@ -328,6 +327,37 @@ export class ParserIncludes {
...
+
+                const gitCloneBranch = (ref === "HEAD") ? "" : `--branch ${ref}`;
+                await Utils.bashMulti([
+                    `cd ${cwd}/${stateDir}`,
+                    `git clone ${gitCloneBranch} -n --depth=1 --filter=tree:0 ${remote.schema}://${remote.host}:${remote.port}/${project}.git ${cwd}/${target}.${ext}`,
+                    `cd ${cwd}/${target}.${ext}`,
+                    `git sparse-checkout set --no-cone ${files[0]} ${files[1]}`,
+                    "git checkout",
+                    `cd ${cwd}/${stateDir}`,
+                    `rsync -a --ignore-missing-args ${cwd}/${target}.${ext}/templates ${cwd}/${target}`, // use rsnyc to ignore missing templates/, the check for existence is in a later step
```
ANGkeith: i see that this is inspired from `downloadIncludeProjectFile` , wondering if we should clean up the temporary folder (`${cwd}/${target}.${ext}`) 

----

under what the scenario will `--ignore-missing-args` be required ?

src/parser-includes.ts
```diff
@@ -328,6 +327,37 @@ export class ParserIncludes {
...
+                ], cwd);
+            } else {
+                await fs.mkdirp(`${cwd}/${target}`);
+                await Utils.bash(`set -eou pipefail; git archive --remote=ssh://git@${remote.host}:${remote.port}/${project}.git ${ref} ${files[0]} ${files[1]} | tar -f - -xC ${target}/`, cwd);
```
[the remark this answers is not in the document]
kk-chesse: Thank you for testing ssh. This is something I couldn't do, because our instance doesn't have ssh :(

I think we can work around this by using wildcards `templates/component*.yml`. This works at least on local repositories. I pushed the wild card version could you test it?

src/parser-includes.ts
kk-chesse: Sure, I added the cleanup code to both functions

Thank you for your review!

kk-chesse:
Sorry forgot to run the linter. Im not using TS often.

ANGkeith:
close #1700

PR #1718, opened 2025-12-21
Convert to bun

firecow:


firecow:
@ANGkeith It actually works. The tests are exploding, though 😄

ANGkeith:
will be curious if we'll be getting any performance gain from this migration

PR #1774, opened 2026-02-22
Make built binaries compatible with pre-AVX2 CPUs

galkinvv:
Use `-baseline` variant for targets supporting it based on
https://bun.com/docs/bundler/executables#supported-targets

---
## Summary by cubic
Build x64 binaries with Bun’s -baseline targets so they run on CPUs without AVX2. Updates Linux x64, macOS x64, and Windows x64 builds; ARM and Node builds unchanged.

<sup>Written for commit bf12f7be5f70fbd977143fb3e2785102a70fe6d9. Summary will update on new commits.</sup>


galkinvv:
The bun-built 4.67 linux gitlab-ci-local binary crashes on start with `illegal hardware instruction` when running on pre-AVX2 Sandy Bridge-EP Workstation (LGA 2011, Xeon E5 2687W). The crash happens even when running without any arguments in folder without `.gitlab-ci.yml`

Thats a regression compared to 4.65.1 - it worked fine.

gdb discovers that it crashes on `rorx` instruction introduced in Bit Manipulation Instruction Set 2 (synchronously with AVX2).

I've tried updated buns' target to the `-baseline` variant - and it actually helps. Here are [example binaries](https://github.com/galkinvv/gitlab-ci-local/actions/runs/22283418961) from an extra temporal branch that releases binaries as artifacts .zip

Those example binaries work on my workstation - local CI is executed fine. I've noted that using `objdump -d` the resulting binary still show some `rorx` instructions left, maybe they are for some runtime-detected code paths.

I understand that the CPU is 14 years old, and supporting it may have negative effects in required efforts and performance (well, gitlab-ci-local itself is extremely fast on all hardware I seen, so this maybe a bit theoretical).

So this PR is a suggestion to support it, but I'd understand its rejection. Anyway, thanks for supporting this useful tool! (I'll continue to update it on my newer PC, and just freeze the version on the older one)

firecow:
It trades a theoretical micro-performance gain for broader hardware compatibility. 

I say yes 👍 We might revert if modern users are impacted by this, but there is only one way to really find out.

We will SHIP it !!! 😆

firecow:
@galkinvv New version will be out soon.
https://github.com/firecow/gitlab-ci-local/releases/tag/4.67.1

galkinvv:
> @galkinvv New version will be out soon. https://github.com/firecow/gitlab-ci-local/releases/tag/4.67.1

Tested this release on same machine - works fine! Thanks.

PR #1778, opened 2026-02-26
Fix GCL_ array env vars not splitting semicolon-separated values

### LINKED ISSUE #992: Error initializing Terraform backend
```
.terraform-format:
  stage: terraform-validate
  script:
    - terraform init
    - terraform fmt -check -diff -recursive
```

**Expected behavior**
When I run gitlab-ci-local, Terraform fails to initialise against an AWS S3 backend (bucket).
Normally this works fine when AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are set correctly as env vars. (or in Gitlab itself in the CI CD Variables section).
I have tried setting them in GLC_VARIABLES but that has no effect.

**Host information**
MacOS
gitlab-ci-local 4.43.0

**Containerd binary**
Docker

**BTW**
Works fine when I put them into .gitlab-ci-local-variables but would like to use ENV VAR GLC_VARIABLES as I work with dotenv a lot so I would not need to define them in different locations.

firecow:
- Fix `GCL_VARIABLE` and other array-type env vars not splitting semicolon-separated values (e.g. `GCL_VARIABLE="VAR1=hello;VAR2=world"`)
- Array keys derived from yargs at runtime — no hardcoded list to maintain
- Simplify `argv.ts`: `getStringArray()` helper reduces getter boilerplate, early return in `injectDotenv()`
- Fix regex backtracking vulnerability in variable getter (SonarCloud S5852)

Fixes #992


cedric-orange:
Do you have any idea when the merge will occurs ?

firecow:
@cedric-orange During easter 😄

PR #1815, opened 2026-03-25
feat: new flag --list-rule

clarson2022:
### Context
In my use cases, we have complex pipeline configurations and having the ability to see what rule is being matched is helpful for debugging when jobs unexpectedly are or aren't being added to the pipeline.

### Proposed Solution
This new flag would result in similar output to --list but would display the rule that a job matched on, i.e. the one that was triggered by rules conditions that result in it being added to the "list". 

#### Changes
- main changes are to runList(): in the case that listRule is set and a job has rules, rather than displaying the 'needs' of the job, display the rule it matched on. 
- other changes are to support capturing the matching rule when getRulesResult determines it
- test case list-rule-case added with 2 tests

---
## Summary by cubic
Adds a new `--list-rule` flag to list jobs and show which rule condition(s) matched (if/exists/changes), replacing the needs column in this mode. Works with `--evaluate-rule-changes` to surface matched `changes` patterns.

- **New Features**
  - New CLI flag `--list-rule` (+ help text) that switches list view to a “rule” column.
  - Prints matched condition(s) from `Utils.getRulesResult` into `Job.matchedRule`; rules with only `when` do not print.
  - README updated with examples; added integration tests for if, exists, changes, combined cases, and when-only behavior.

- **Bug Fixes**
  - Updated `rules.test.ts` to assert `matchedRule` outputs; fixed ESLint and yamllint issues.

<sup>Written for commit 93766dee53e70f458b6e0c359045d86f7c550cbc. Summary will update on new commits.</sup>


firecow:
Feel free to reopen if you get time to finalize this

PR #1825, opened 2026-04-05
fix: pass artifact paths to rsync via `--files-from` 

Paul-Goulpie:
## fix: pass artifact paths to rsync via `--files-from`                                                                                                                                                          
                                                                                                                                                                                                                 
### Problem                                                                                                                                                                                                      
                                                                                                                                                                                                                 
When copying artifacts out, each path listed under `artifacts.paths`                                                                                                                                             
was appended directly as a rsync CLI argument. With a large number of
files (or glob patterns expanding to many files), the total size of                                                                                                                                              
the argument list could exceed `ARG_MAX` (~2 MB on Linux), causing                                                                                                                                               
rsync to fail silently (swallowed by `|| true`) and producing no                                                                                                                                                 
artifacts.                                                                                                                                                                                                       
                                                                                                                                                                                                                 
Additionally, glob patterns such as `artifact_*` were passed                                                                                                                                                     
single-quoted to rsync, preventing bash from expanding them — so
rsync received a literal `artifact_*` and matched nothing.                                                                                                                                                       
                                                                                                                                                                                                                 
### Solution                                                                                                                                                                                                     
                                                                                                                                                                                                                 
- Collect artifact paths into a temporary file using a bash `for`                                                                                                                                                
  loop (a shell built-in, not subject to `ARG_MAX`)
- The `for` loop correctly expands glob patterns before writing                                                                                                                                                  
  individual resolved paths to the temp file                    
- Pass the temp file to rsync via `--files-from` so the argument
  list remains constant (3 args) regardless of the number of files
- Remove the temp file after rsync completes                                                                                                                                                                     
 
### Test                                                                                                                                                                                                         
                                                                
Added `tests/test-cases/artifacts-many-paths`: creates 9 000 files                                                                                                                                               
with 255-character names (~2.3 MB of paths, exceeding `ARG_MAX`),
exports them via a glob pattern, and asserts in the consume stage
that all 9 000 files were transferred correctly.

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Fix artifact export by switching to rsync `--files-from` with a temp file to avoid ARG_MAX failures and correctly expand globs. This makes large artifact sets transfer reliably.

- **Bug Fixes**
  - Use `mktemp` and a bash `for` loop with `printf` to expand globs and write each path to a temp file.
  - Run `rsync -rRa --files-from` to copy files/dirs while honoring excludes, then remove the temp file; add an integration test with 9,000 long-named files under `--shell-isolation` asserting “exported artifacts” and full consumption.

<sup>Written for commit ddbd777ae65dbccc71bbb9113491866d0fa24ad9. Summary will update on new commits.</sup>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/src/job.ts
+++ b/src/job.ts
@@ -1424,16 +1424,18 @@ If you know what you're doing and would like to suppress this warning, use one o
         let time, endTime;
         let cpCmd = "shopt -s globstar nullglob dotglob\n";
         cpCmd += `mkdir -p ${artifactsPath}/${safeJobName}\n`;
-        cpCmd += "rsync --exclude '.gitlab-ci-local/**' -Ra ";
+        cpCmd += "_gcl_files_tmp=\\$(mktemp)\n";
+        for (const artifactPath of this.artifacts?.paths ?? []) {
+            const expandedPath = Utils.expandText(artifactPath, expanded).replace(`${expanded.CI_PROJECT_DIR}/`, "");
+            cpCmd += `for _gcl_f in ./${expandedPath}; do printf '%s\\n' "\\$_gcl_f"; done >> \\$_gcl_files_tmp\n`;
+        }
+        cpCmd += "rsync --exclude '.gitlab-ci-local/**' -rRa ";
         for (const artifactExcludePath of this.artifacts?.exclude ?? []) {
             const expandedPath = Utils.expandText(artifactExcludePath, expanded).replace(`${expanded.CI_PROJECT_DIR}/`, "");
             cpCmd += `--exclude '${expandedPath}' `;
         }
-        for (const artifactPath of this.artifacts?.paths ?? []) {
-            const expandedPath = Utils.expandText(artifactPath, expanded).replace(`${expanded.CI_PROJECT_DIR}/`, "");
-            cpCmd += `./${expandedPath} `;
-        }
-        cpCmd += `${artifactsPath}/${safeJobName}/. || true\n`;
+        cpCmd += `--files-from=\\$_gcl_files_tmp . ${artifactsPath}/${safeJobName}/. || true\n`;
+        cpCmd += "rm -f \\$_gcl_files_tmp\n";
         const reportDotenv = Utils.expandText(this.artifacts.reports?.dotenv ?? null, expanded);
         const reportDotenvs: string[] | null = (typeof reportDotenv === "string") ? // normalize to string[] for easier handling
             [reportDotenv] :
--- a/tests/test-cases/artifacts-many-paths/.gitlab-ci.yml
+++ b/tests/test-cases/artifacts-many-paths/.gitlab-ci.yml
@@ -1,0 +1,30 @@
+---
+# Creates 9000 files whose names are 255 characters long.
+# With the old approach (paths passed as rsync arguments), the total argument
+# size (~2.3 MB) exceeded ARG_MAX (~2 MB) and rsync failed silently.
+# With the new approach (--files-from + bash for loop), there is no limit on
+# the number of files.
+produce-artifacts:
+  stage: build
+  script:
+    - |
+      python3 -c "
+      import os
+      padding = 'x' * 240
+      for i in range(9000):
+          name = f'artifact_{i:05d}_{padding}'
+          fd = os.open(name, os.O_CREAT | os.O_WRONLY, 0o644)
+          os.close(fd)
+      "
+    - mkdir -p output_dir && echo "hello" > output_dir/result.txt
+  artifacts:
+    paths:
+      - artifact_*
+      - output_dir/
+
+consume-artifacts:
+  stage: test
+  script:
+    - count=$(find . -maxdepth 1 -name 'artifact_*' | wc -l)
+    - if [ "$count" -ne 9000 ]; then echo "FAIL expected 9000 artifacts, got $count"; exit 1; fi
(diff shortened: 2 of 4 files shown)


src/job.ts
```diff
@@ -1424,16 +1424,18 @@ If you know what you're doing and would like to suppress this warning, use one o
         let time, endTime;
         let cpCmd = "shopt -s globstar nullglob dotglob\n";
         cpCmd += `mkdir -p ${artifactsPath}/${safeJobName}\n`;
+        cpCmd += "_gcl_files_tmp=\\$(mktemp)\n";
+        for (const artifactPath of this.artifacts?.paths ?? []) {
+            const expandedPath = Utils.expandText(artifactPath, expanded).replace(`${expanded.CI_PROJECT_DIR}/`, "");
+            cpCmd += `for _gcl_f in ./${expandedPath}; do if [ -d \\$_gcl_f ]; then find \\$_gcl_f -print >> \\$_gcl_files_tmp; else printf '%s\\n' \\$_gcl_f >> \\$_gcl_files_tmp; fi; done\n`;
```
firecow, line 1430: `--files-from` suppresses `-r` from `-a`, even when `-a` normally implies it. The man page states:

> *"The --archive (-a) option's behavior does not imply --recursive (-r), so specify it explicitly, if you want it."*

The PR compensates with `if [ -d ]; then find ... -print`, but this has two downsides:

1. **Double traversal** — `find` walks the entire tree to write every path to the temp file, then the sync tool reads that list and resolves each path individually. Its internal directory walker is heavily optimized; reimplementing it via `find` through a text file is strictly worse.

2. **Bloated temp file** — a directory with 100k files produces a 100k-line temp file. With explicit `-r`, the temp file only contains the top-level glob results (e.g. `./dir`) and stays small regardless of tree depth.

Simpler fix: add explicit `-r` to the invocation and drop the `find` branching entirely:

```typescript
cpCmd += "_gcl_files_tmp=\\$(mktemp)\n";
for (const artifactPath of this.artifacts?.paths ?? []) {
    const expandedPath = Utils.expandText(artifactPath, expanded).replace(`${expanded.CI_PROJECT_DIR}/`, "");
    cpCmd += `for _gcl_f in ./${expandedPath}; do printf '%s\\n' "\\$_gcl_f"; done >> \\$_gcl_files_tmp\n`;
}
cpCmd += "rsync --exclude '.gitlab-ci-local/**' -rRa ";
```

Same ARG_MAX fix, same glob expansion fix, fewer moving parts, better performance on large directory trees.

firecow, line 1430: The redirect `>>` is inside the loop body, so the temp file is opened and closed on every iteration:

```bash
for _gcl_f in ./pattern; do printf '%s\n' $_gcl_f >> $tmpfile; done
```

Move the redirect outside the loop so it opens the file once per artifact pattern:

```bash
for _gcl_f in ./pattern; do printf '%s\n' "$_gcl_f"; done >> $tmpfile
```

For the 9000-file test case this is 9000 open/close cycles vs 1. For larger projects it becomes a real bottleneck.

firecow, line 1430: `$_gcl_f` and `$_gcl_files_tmp` are unquoted throughout. Paths with spaces break `[ -d $var ]`, `find $var`, and `printf '%s\n' $var` due to word splitting and glob re-expansion. This is pre-existing (the old code had the same issue), but since the code is being rewritten it's worth fixing:

```bash
printf '%s\n' "$_gcl_f"
```

tests/test-cases/artifacts-many-paths/.gitlab-ci.yml
```diff
@@ -0,0 +1,27 @@
...
+      padding = 'x' * 240
+      for i in range(9000):
+          name = f'artifact_{i:05d}_{padding}'
+          fd = os.open(name, os.O_CREAT | os.O_WRONLY, 0o644)
+          os.close(fd)
+      "
+  artifacts:
+    paths:
+      - artifact_*
+
```
firecow, line 22: Good test — it genuinely exceeds ARG_MAX and catches the silent `|| true` failure. Consider also adding an artifact path that resolves to a directory (not just a glob of flat files), since that's where `--files-from` changes the sync semantics:

```yaml
artifacts:
  paths:
    - artifact_*
    - output_dir/
```

This would guard against the `-r` suppression specifically.

firecow (changes requested):
The `--files-from` approach for ARG_MAX is the right idea. A few issues to address — the main one being that `--files-from` silently suppresses `-r` from `-a`, which the PR works around with `find`, but adding explicit `-r` to the sync command is simpler and more performant.

Paul-Goulpie:
@firecow 
Thanks for the detailed review — all four points addressed:                                                                                                                                                      
                                                                                                                                                                                                                 
- Dropped find branching, added explicit -r: The loop now uses a simple printf '%s\n' "$_gcl_f" and rsync is invoked with -rRa. Verified locally that -r + --files-from correctly recurses into directory        
entries.                                                                                                                                                                                                         
- Redirect moved outside the loop: done >> \$_gcl_files_tmp so the file is opened once per artifact pattern instead of once per file.
- Variables quoted: "$_gcl_f" throughout.                                                                                                                                                                        
- Test enhanced with a directory artifact: Added output_dir/ to the artifact paths in .gitlab-ci.yml and a corresponding assertion in consume-artifacts, specifically to guard against -r suppression semantics.
                                                                                                                                                                                                                 
Full test suite passes (147/147).

firecow (approved):

PR #1836, opened 2026-04-20
Allow to use `component.reference`, `component.sha`, `component.version` and `component.name`

jrd:
When including another component B from a component A, allow to use `$[[ component.reference ]]` notation for component B.

### DIFF
--- a/src/parser-includes.ts
+++ b/src/parser-includes.ts
@@ -29,8 +29,12 @@ type ParsedComponent = {
     domain: string;
     port: string;
     projectPath: string;
+    componentPath: string;
+    effectiveRef: string;
     name: string;
-    ref: string;
+    reference: string;
+    version: string | undefined;
+    sha: string;
     isLocal: boolean;
 };
 
@@ -96,11 +100,11 @@ export class ParserIncludes {
             } else if (value["remote"]) {
                 promises.push(this.downloadIncludeRemote(cwd, stateDir, value["remote"], fetchIncludes, writeStreams));
             } else if (value["component"]) {
-                const component = this.parseIncludeComponent(value["component"], gitData);
+                const component = this.parseIncludeComponent(value["component"], gitData, opts);
                 componentParseCache.set(index, component);
                 if (!component.isLocal)
                 {
-                    promises.push(this.downloadIncludeComponent(opts, component.projectPath, component.ref, component.name));
+                    promises.push(this.downloadIncludeComponent(opts, component.projectPath, component.effectiveRef, component.componentPath));
                 }
             }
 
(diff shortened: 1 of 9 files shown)


firecow:
Apply cubic's `component\.` suggestion — it's still open. And drop the `compList.includes(...)` check in `getComponentValue`; upstream GitLab doesn't require declaring `spec.component`, so just always allow `reference`, `version`, `sha`.

jrd:
I updated my pull request, also borrowing the better idea from #1839

It should be all good now.

src/parser-includes.ts
```diff
@@ -458,3 +458,53 @@ export async function resolveIncludeLocal (pattern: string, cwd: string) {
...
+    // effectiveRef may already be a sha, if so return it directly
+    if (/^[0-9a-f]{40}$/.test(effectiveRef)) {
+        return effectiveRef;
+    }
+    const stdout = getGitRemoteInfo(ctx);
+    const match = stdout.split("\n").find(line =>
+        line.endsWith(`refs/tags/${effectiveRef}^{} `) ||
+        line.endsWith(`refs/tags/${effectiveRef}`) ||
+        line.endsWith(`refs/heads/${effectiveRef}`),
+    );
```
firecow: For annotated tags this returns the wrong SHA. The `^{} ` pattern has a stray trailing space so it never matches, and `.find` hits `refs/tags/x` (the tag-object SHA) before `refs/tags/x^{}` (the commit SHA). Prefer the deref'd line:

```suggestion
    const lines = stdout.split("\n");
    // annotated tags: prefer the deref'd commit sha (refs/tags/x^{})
    const match = lines.find(line => line.endsWith(`refs/tags/${effectiveRef}^{}`))
        ?? lines.find(line =>
            line.endsWith(`refs/tags/${effectiveRef}`) ||
            line.endsWith(`refs/heads/${effectiveRef}`));
```

firecow:
One behavior note (not blocking): `getComponentSha` runs a full `git ls-remote` on every remote component include, even when the template never uses `$[[ component.sha ]]`. Previously only semver ranges hit the network. Could the SHA be resolved lazily — only when the placeholder is actually present?

jrd:
> One behavior note (not blocking): `getComponentSha` runs a full `git ls-remote` on every remote component include, even when the template never uses `$[[ component.sha ]]`. Previously only semver ranges hit the network. Could the SHA be resolved lazily — only when the placeholder is actually present?

Yes, a lazy version is preferable in that case. I see what I can manage to do it in typescript. If I can find a better solution, a lambda/arrow function will do it.

[Edit] I think typescript have property, at least there is getter function in javascript, that could do the trick to compute it only when required.

jrd:
@firecow Do you want me to squash the last two commits with the first one?

firecow:
> @firecow Do you want me to squash the last two commits with the first one?

Nah, don't care.. I'll squash into master anyway

jrd:
So… what’s missing for merge?

jrd:
Why no answer? Why not merge this?

firecow (approved):

PR #1595, opened 2025-06-24
Support image.docker.platform

harryjms:
Adds support for `image.docker.platform` in the yaml

PR #1138, opened 2024-03-12
fix(#1137): only add image.docker.user if the value exists

e-picas:
This fixes #1137

With the original YML in the ticket:

```yaml
my_job:
  image: node:lts
  script:
    - echo 'test'
```
The rendering of the `--preview` option is now safe (no more `image.docker.user: null`):

```bash
$ node ../test-gitlab-ci-local/gitlab-ci-local/src/index.js --preview
---
stages:
  - .pre
  - build
  - test
  - deploy
  - .post
my_job:
  image:
    name: node:lts
  script:
    - echo 'test'
```
And it passes the `ci/lint` validation:

```bash
$ jq --null-input --arg yaml "$(node ../test-gitlab-ci-local/gitlab-ci-local/src/index.js --preview)" '.content=$yaml' \
| curl "https://gitlab.com/api/v4/projects/${PROJECT_ID}/ci/lint" \
	--header "Authorization: Bearer ${GITLAB_TOKEN}" \
	--header 'Content-Type: application/json' \
	--data @- | jq '.errors'

[]
```

I also deleted the `image.docker.user: null` in the test cases.

PR #1856, opened 2026-05-09
feat: support image.docker.platform

inistor:
## Summary

Adds support for `image.docker.platform` per the GitLab CI YAML reference ([introduced in GitLab 18.0](https://docs.gitlab.com/ci/yaml/#imagedockerplatform)). The platform string is forwarded to:

- `docker pull --platform <platform>` — fetch the requested manifest
- `docker run --platform <platform>` — guard against a multi-arch image already cached locally with the wrong default

The existing `pulled` log line is suffixed with `(<platform>)` when set, e.g. `pulled alpine (linux/amd64) in 2.3 s`. This gives the user feedback about what was actually pulled, and gives the integration test a concrete assertion surface.

## Relationship to #1595

This PR supersedes the abandoned attempt in #1595 (closed Feb 17 2026 with "*Feel free to reopen, once eslint has been fixed and tests have been added*"). Same goal, but:

1. **ESLint clean** — the eight `import { X }` style violations flagged inline by @bbrala on the original PR are fixed; all imports use `import {X}` (no inner spaces).
2. **Tests added** — new `image-platform` fixture + integration test in `tests/test-cases/image/`.

A few small design tweaks vs #1595:

- **No new "pulling" log line.** #1595 emitted both `pulling` (new) and `pulled` (existing). I just suffix `(<platform>)` onto the existing `pulled` line. Avoids changing test snapshots for every other test that pulls an image.
- **Reuses the local `imagePlatform`** in the run-cmd path instead of calling `this.imagePlatform(expanded)` twice.
- **`pullImage` signature change is additive** (`imagePlatform: string | null = null`); the helper-image and service-pull callsites don't need touching.

## Implementation

| File | Change |
|---|---|
| `src/data-expander.ts` | `imageComplex` propagates `docker.platform` alongside `docker.user`. Preserves the spirit of #1138 (don't emit null fields) — verified via `--preview` that `image-user` and `image-platform` fixtures both render without spurious null keys. |
| `src/job.ts` | New `imagePlatform()` private method (mirrors `imageUser()`); threaded through `pullImage` to add `--platform` to the pull spawn; same `--platform` appended to the `docker run` cmd; `pulled` log suffix. |
| `tests/test-cases/image/.gitlab-ci.yml` | Adds `image-platform` job using `linux/amd64` (matches CI runner arch). |
| `tests/test-cases/image/integration.test.ts` | New integration test: `pullPolicy: "always"` to force a fresh pull, asserts the `pulled` log line includes `(linux/amd64)`, asserts `uname -m` outputs `x86_64`. |

## Real-world validation

Beyond the integration test, validated against a private project that uses `image.docker.platform: linux/${PROCESSOR}` together with `parallel.matrix: PROCESSOR: [x86_64, aarch64]` and a `before_script` that asserts `[ "$PROCESSOR" = "$(uname -m)" ]`:

- **Released `gitlab-ci-local` 4.71.0** + `[x86_64]` permutation → ❌ `Mismatch between host processor (aarch64) and target processor (x86_64)` (the `docker.platform` field is silently dropped during YAML expansion, so docker pulls the host-arch manifest).
- **This branch** + `[x86_64]` → ✅ PASS, full library compile under qemu-x86_64 emulation on an Apple-Silicon host (39 s).
- **This branch** + `[aarch64]` → ✅ PASS, native (3.6 s).

The 10× wall-time difference is itself proof that x86_64 is genuinely running under emulation, not on the aarch64 host.

## Out of scope (follow-up if requested)

- `services[].docker.platform` exists in the same GitLab 18.0 keyword family. Adding it cleanly would also require `servicesComplex` to start propagating `docker.*` (it currently strips it entirely — there's no `services[].docker.user` either). Sibling PR.

## References

- Issues: closed PR firecow/gitlab-ci-local#1595, PR firecow/gitlab-ci-local#1138 (the "only emit if value exists" precedent for `image.docker.user`)
- GitLab docs: [`image:docker:platform`](https://docs.gitlab.com/ci/yaml/#imagedockerplatform), [`services:docker:platform`](https://docs.gitlab.com/ci/yaml/#servicesdocker)
- Original implementation of `image.docker.user`: firecow/gitlab-ci-local#1130

## Test plan

- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bunx vitest run tests/test-cases/image/` (10/10 passed)
- [x] CI green on internal fork PR before opening this one ([inistor#2](https://github.com/inistor/gitlab-ci-local/pull/2) — all 11 checks passed including vitest, smoke-test × 4, CodeQL, eslint, typecheck, yamllint, unused-deps)

## Acknowledgment

This PR was implemented with assistance from Claude (Anthropic) as a pair-programming tool. The implementation was audited against the GitLab CI YAML spec and existing patterns in this codebase (notably `image.docker.user` from #1130, the null-emission fix in #1138, and the eight import-style review comments on #1595). All code and tests were reviewed by me before pushing; CI on the internal fork PR confirmed the full suite passes. Co-authorship is preserved in the commit trailer.

---
## Summary by cubic
Adds support for `image.docker.platform` (GitLab 18.0) and ensures the requested architecture is always used by forwarding the platform to Docker and reflecting it in logs. Also forces a pull when a platform is set to avoid cache mismatches.

- **New Features**
  - Forwards `image.docker.platform` to `docker pull` and `docker run` via `--platform <value>`.
  - Adds the platform to the `pulled` log line, e.g. `pulled alpine (linux/amd64) in 2.3 s`.
  - Emits `docker.platform` only when set during YAML expansion.
  - Integration test (`image-platform`) verifies the log output and `uname -m` on `linux/amd64` without `pullPolicy: "always"`.

- **Bug Fixes**
  - When a platform is set, bypasses the platform-agnostic `image inspect` cache check and always runs `docker pull` to prevent wrong-arch variants from being used.

<sup>Written for commit 5fc70f4aeb3b42aa02ca9f398d7b43ecfa9586c9. Summary will update on new commits.</sup>

### DIFF
--- a/src/data-expander.ts
+++ b/src/data-expander.ts
@@ -188,7 +188,7 @@ export function imageComplex (data: any) {
     return {
         name: typeof data === "string" ? data : data.name,
         entrypoint: data.entrypoint,
-        ...(data.docker?.user ? {docker: {user: data.docker?.user}} : {}),
+        ...(data.docker?.user || data.docker?.platform ? {docker: {user: data.docker.user, platform: data.docker.platform}} : {}),
     };
 }
 
--- a/src/job.ts
+++ b/src/job.ts
@@ -670,6 +670,7 @@ If you know what you're doing and would like to suppress this warning, use one o
         this._dotenvVariables = await this.initProducerReportsDotenvVariables(writeStreams, Utils.expandVariables(this._variables));
         const expanded = Utils.unscape$$Variables(Utils.expandVariables({...this._variables, ...this._dotenvVariables}));
         const imageName = this.imageName(expanded);
+        const imagePlatform = this.imagePlatform(expanded);
         const helperImageName = argv.helperImage;
         const safeJobName = this.safeJobName;
 
@@ -682,7 +683,7 @@ If you know what you're doing and would like to suppress this warning, use one o
         }
 
         if (imageName) {
-            await this.pullImage(writeStreams, imageName);
+            await this.pullImage(writeStreams, imageName, imagePlatform);
 
             const buildVolumeName = this.buildVolumeName;
             const tmpVolumeName = this.tmpVolumeName;
@@ -971,6 +972,11 @@ If you know what you're doing and would like to suppress this warning, use one o
                 dockerCmd += `--user ${imageUser} `;
             }
 
+            const imagePlatform = this.imagePlatform(expanded);
+            if (imagePlatform) {
+                dockerCmd += `--platform ${imagePlatform} `;
+            }
+
             if (this.argv.containerEmulate) {
                 const runnerName: string = this.argv.containerEmulate;
 
@@ -1207,6 +1213,13 @@ If you know what you're doing and would like to suppress this warning, use one o
         return Utils.expandText(image["docker"]["user"], vars);
     }
 
+    private imagePlatform (vars: {[key: string]: string} = {}): string | null {
+        const image = this.jobData["image"];
+        if (!image) return null;
+        if (!image["docker"]) return null;
+        return Utils.expandText(image["docker"]["platform"], vars);
+    }
+
     get imageEntrypoint (): string[] | null {
         const image = this.jobData["image"];
 
@@ -1235,14 +1248,16 @@ If you know what you're doing and would like to suppress this warning, use one o
         }
     }
 
-    private async pullImage (writeStreams: WriteStreams, imageToPull: string) {
+    private async pullImage (writeStreams: WriteStreams, imageToPull: string, imagePlatform: string | null = null) {
         const pullPolicy = this.argv.pullPolicy;
+        const platformArgs = imagePlatform ? ["--platform", imagePlatform] : [];
+        const platformSuffix = imagePlatform ? ` (${imagePlatform})` : "";
         const actualPull = async () => {
             await this.validateCiDependencyProxyServerAuthentication(imageToPull);
             const time = process.hrtime();
-            await Utils.spawn([this.argv.containerExecutable, "pull", imageToPull]);
+            await Utils.spawn([this.argv.containerExecutable, "pull", imageToPull, ...platformArgs]);
             const endTime = process.hrtime(time);
-            writeStreams.stdout(chalk`${this.formattedJobName} {magentaBright pulled} ${imageToPull} in {magenta ${prettyHrtime(endTime)}}\n`);
+            writeStreams.stdout(chalk`${this.formattedJobName} {magentaBright pulled} ${imageToPull}${platformSuffix} in {magenta ${prettyHrtime(endTime)}}\n`);
             this.refreshLongRunningSilentTimeout(writeStreams);
         };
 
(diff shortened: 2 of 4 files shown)


firecow:
Isn't docker run on a ARM64 machine using ARM64 docker images on its own or what is it that I'm not getting here?

inistor:
> Isn't docker run on a ARM64 machine using ARM64 docker images on its own or what is it that I'm not getting here?

Indeed, docker run will use the host's architecture, unless otherwise requested:
```
docker run -it --rm ubuntu uname -a
Linux a03fd021bff7 6.12.76-linuxkit #1 SMP Thu Apr 30 15:11:03 UTC 2026 aarch64 GNU/Linux
```
However, if you're interested in a specific arch, you need to request it explicitly:
```
➜  ~ docker run -it --rm --platform linux/aarch64 ubuntu uname -a
Linux 8b22cc799367 6.12.76-linuxkit #1 SMP Thu Apr 30 15:11:03 UTC 2026 aarch64 GNU/Linux
➜  ~ docker run -it --rm --platform linux/amd64 ubuntu uname -a
Linux 74ac3b7bdb48 6.12.76-linuxkit #1 SMP Thu Apr 30 15:11:03 UTC 2026 x86_64 GNU/Linux
```

For my cross-compiling case, let's take the following example (my production runners are mainly x86_64 - but we also have some arm64s - but the local env is Mac - aarch64). The recipe looks like this:
```
.linux-template:
  tags:
    - docker-${PROCESSOR}
  image:
    name: my-builder:latest
    docker:
      platform: linux/${PROCESSOR}
  parallel:
    matrix:
      - PROCESSOR: [x86_64, aarch64]
  before_script:
    # Guard: make sure we're actually running on the requested arch.
    # Without image.docker.platform, both matrix permutations resolve to the
    # host's native arch and one of them trips this assertion.
    - if [ "$PROCESSOR" != "$(uname -m)" ]; then
        echo "Mismatch between host processor ($(uname -m)) and target processor ($PROCESSOR)";
        exit 1;
      fi
build-mylib:
  extends: .linux-template
  script:
    - ./build.sh --arch=$PROCESSOR
```

Now, if for some reason, the wrong arch runner would pick it up (of course, we do have the "tags", selecting the arch, so in production this should not theoretically happen), you would end-up with a job "named" aarch64 but actually being x86_64 (it actually happened at some point, due to some tag confusion - and it's quite hard to figure out, as you can imaging). This way (with the `platform` selector), at least, if the runner is incorrectly selected, it would trigger a cross-platform build; much much slower, but at least correct.

Similarly, debugging with  `gitlab-ci-local` (since tag is irrelevant in this case) would by default use `aarch64` - despite the fact that the intent is `x86_64`.

Not sure if the above answers your question - I'll be more than happy to provide more clarifications if needed.

bbrala:
Yeah im running into this also. Some of the images i need to run against don't have builds for all architectures so i want to force amd64 and let the virtualization handle things.

firecow (approved):

inistor:
Thanks @firecow !

Issue #1859, opened 2026-05-12
regression between `4.71.0` and `4.72.0`


ticapix:
Hello,

As hinted in https://github.com/firecow/gitlab-ci-local/issues/1844#issuecomment-4404800772, I experience a regression bug that I reproduced here

https://gitlab.com/pierre-gronlier/debug-ci with the `./test.sh` which dl version 4.71.0 and 4.72.0 and run it against:

- the TBC semver template
- a simplified version of the template

**Minimal .gitlab-ci.yml illustrating the issue**
```yml
workflow:
  rules:
    - if: $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME =~ $PROD_REF || $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME =~ $INTEG_REF
      when: never
variables:
  PROD_REF: /^(master|main)$/
```

The error is
```
  Error attempting to evaluate the following rules:
    rules:
      - if: 'null =~ "/^(master|main)$/" || null =~ null'
  as
  ```javascript
  false || null =~ null
  ```
```

ticapix:
using bisect script, I narrowed it down to commit https://github.com/firecow/gitlab-ci-local/commit/fcccb7ecf715efc151b42ff8c7379ab013b432b7

--- Claude analysis ---

First bad commit: fcccb7e — "feat: implement workflow:rules:variables support (#1832) (#1833)"

  What the commit does

  It adds a new code path in src/parser.ts that evaluates workflow:rules early during parsing to extract and merge variables into the global scope.

  Why it causes the bug

  The evaluation calls Utils.getRulesResult with the workflow rules and a variable set that does not include $INTEG_REF (it's undefined at parse time). The rule expression is:

  $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME =~ $PROD_REF || $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME =~ $INTEG_REF

  With $CI_MERGE_REQUEST_SOURCE_BRANCH_NAME = null (not in MR context) and $INTEG_REF = null (undefined), this reduces to:

  false || null =~ null   // ← throws instead of returning false

  The null =~ null case in the rules evaluator throws an error rather than gracefully returning false.

  Why it didn't fail before

  Before fcccb7e, workflow rules were not evaluated during --preview parsing. The new code path is what exposes the pre-existing null =~ null unhandled case in the expression evaluator.

  The fix

[...]

PR #1860, opened 2026-05-12
fix(rules): treat null RHS in =~ and !~ as false

ticapix:
fix for #1859 

When a variable used as the RHS of a regex match is undefined, GitLab CI evaluates the expression as null (falsy) for both =~ and !~. Previously gitlab-ci-local would throw or produce incorrect results. Replace such expressions with "false" before evaluation.

---
## Summary by cubic
Fixes rule evaluation when the RHS of =~ or !~ is undefined by treating it as false, matching GitLab CI and preventing crashes or wrong results. Affects workflow rule conditions that use regex with missing variables.

- **Bug Fixes**
  - Replace patterns like '... =~ null' and '... !~ null' with 'false' before evaluation.
  - Added `.gitlab-ci-null-rhs-regex.yml` and an integration test to verify the job runs and no error is thrown.

<sup>Written for commit 4fc232a86a980167db5d2f4cc12606a134c0424e. Summary will update on new commits.</sup>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -299,6 +299,9 @@ export class Utils {
 
         evalStr = evalStr.replaceAll(/null.matchRE2JS\(.+?\)\s*!=\s*null/g, "false");
         evalStr = evalStr.replaceAll(/null.matchRE2JS\(.+?\)\s*==\s*null/g, "true");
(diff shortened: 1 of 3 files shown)


ticapix:
need additional work.

PR #1865, opened 2026-05-22
fix: prevent ENAMETOOLONG crash for long parallel:matrix job names

### LINKED ISSUE #1862: ENAMETOOLONG crash when parallel:matrix generates long job names
# ENAMETOOLONG crash when parallel:matrix generates long job names

## Description

When `parallel:matrix` entries contain long variable values, gitlab-ci-local crashes with `ENAMETOOLONG` before any job executes. The tool cannot run at all for projects that use descriptive matrix parameters.

## Error

```
Error: ENAMETOOLONG: name too long, open '/home/user/project/.gitlab-ci-local/output/my-groupLwcommonLwpython-unit-testOiBbmy-app-controllerLAMyIAappIAcontrollerIAtoIAbeIAusedIAasIAreferenceIAforIAdevelopmentIAteamsLApython311LAcontrollerLAcommonLAcontrollerLwsetupLgpyLAcontrollerLwsetup_cLgpyLAcontrollerLwsetup_nLgpyLAcontrollerLwtestsLyoqLyoscontrollerLwcoverageLyoscontrollerLwbuildLyoqLyoscontrollerLwcoverageLwcoverage-unitLgxmlLA75LAtrueXQ.log'
```

## Steps to Reproduce

`.gitlab-ci.yml`:

```yaml
stages:
  - test

my-group/common/python-unit-test:
  stage: test
  image: alpine:latest
  script:
    - echo "Running unit tests for $module"
  parallel:
    matrix:
      - module: "my-app-controller"
        description: "My app controller to be used as reference for development teams"
        language: "python311"
        folder_path: "controller"
        subfolder_path: "common"
        sonar_exclusions: "controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,controller/coverage/*,controller/build/**/*"
        coverage: "controller/coverage/coverage-unit.xml"
        mutants_max_batch_size: "75"
        pytest_parallel: "true"
```

```bash
gitlab-ci-local --shell-isolation
```

## Expected Behavior

The job should execute normally, as it does on real GitLab CI.

## Actual Behavior

Immediate crash with `ENAMETOOLONG`. No job runs.

## Root Cause

1. **`src/parser.ts:191`** — Job names are built by concatenating ALL matrix variable values:
   ```typescript
   matrixJobName = `${jobName}: [${Object.values(parallelMatrixVariables).join(",")}]`;
   ```

2. **`src/utils.ts:52`** — `safeDockerString` base64url-encodes non-alphanumeric characters, which **expands** the length (e.g., `/` → `Lw`, ` ` → `IA`, `,` → `LA`):
   ```typescript
   static safeDockerString(jobName: string) {
       return jobName.replace(/[^\w-]+/g, (match) => base64url.encode(match));
   }
   ```

3. **`src/job.ts:676`** — The encoded name is used directly as a filename:
   ```typescript
   const outputLogFilePath = `${argv.cwd}/${argv.stateDir}/output/${safeJobName}.log`;
   ```

Linux `NAME_MAX` is 255 bytes. The encoded filename in this case is ~340 characters.

## Why Real GitLab Doesn't Have This Problem

Real GitLab CI identifies jobs by numeric IDs. Matrix variable values are stored in the database and displayed in the UI but never used as filesystem paths.

## Affected Code Paths

`safeJobName` is used for:
- Output logs: `output/${safeJobName}.log`
- Build dirs: `builds/${safeJobName}`
- Docker volumes: `gcl-${safeJobName}-${jobId}-build`
- Service logs: `services-output/${safeJobName}/...`

## Environment

- OS: Linux (ext4/xfs — NAME_MAX=255)
- gitlab-ci-local: latest master
- Also affects macOS (NAME_MAX=255 on APFS/HFS+)

gyanranjan:
## Problem

When `parallel:matrix` entries contain long variable values, `safeDockerString()` encodes non-alphanumeric characters using base64url, which **expands** the string (e.g., `/` → `Lw`, ` ` → `IA`). The encoded name is used directly as a filename/volume name. Linux/macOS `NAME_MAX` is 255 bytes — the reproducer generates ~340 characters, causing an immediate crash.

Fixes #1862

## Fix

When the encoded name exceeds 238 chars, truncate to a readable prefix + `-` + 16-char SHA-256 hash of the original name.

- **Limit:** 238 chars (255 NAME_MAX − 17 worst-case suffix from `gcl-...-999999-build`)
- **Short names:** Unchanged — no hashing when already within limit
- **Deterministic:** Same input always produces same output
- **Unique:** SHA-256 at 64 bits makes collisions negligible
- **Arbitrary length:** Works for any job name length (500, 5000+ chars)

## Changes

- `src/utils.ts`: Added length guard to `safeDockerString()` with hash-based truncation
- `tests/utils.test.ts`: Added 9 test cases covering short names, long names, extreme lengths, boundary conditions, volume name safety, determinism, and uniqueness

## Verification

- `npm run typecheck` ✅
- `npm run lint` ✅
- All 61 unit tests pass ✅
- All 32 parallel-matrix integration tests pass ✅
- Live end-to-end validation with the exact issue reproducer ✅ (was crashing, now passes)
- Two-entry matrix with similar names produces separate files with different hashes ✅
- Short job names produce unchanged output (no regression) ✅

---
## Summary by cubic
Prevents ENAMETOOLONG crashes when `parallel:matrix` generates long job names whose `base64url`-expanded values exceed filesystem limits. Ensures Docker volume/file names stay within 255 bytes. Fixes #1862.

- **Bug Fixes**
  - Derive and enforce `MAX_FILENAME_LENGTH = 255 - 17` (NAME_MAX minus `gcl-${name}-999999-build`) in `safeDockerString`.
  - If over, truncate to a readable prefix and append a 16-char SHA-256 hash of the original; deterministic with low collision; short names unchanged.
  - Added tests: exact base64url assertion, boundaries, extreme lengths, volume safety, and an end-to-end `parallel-matrix-long-name` fixture (YAML updated) to ensure both long and short matrix entries complete.
  - Documented invariants: ASCII-only matching keeps length==bytes; hash the raw name since encoding isn’t injective.

<sup>Written for commit 62e73babe4ed517d2a550426f9987b3172432515. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecow/gitlab-ci-local/pull/1865?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -8,6 +8,7 @@ import checksum from "checksum";
 import base64url from "base64url";
 import execa, {ExecaError} from "execa";
 import assert from "node:assert";
+import {createHash} from "node:crypto";
 import {CICDVariable} from "./variables-from-files.js";
 import {GitData} from "./git-data.js";
 import {globbySync} from "globby";
@@ -49,10 +50,18 @@ export class Utils {
         return url.replace(/^https:\/\//g, "").replace(/^http:\/\//g, "");
     }
 
+    static readonly MAX_FILENAME_LENGTH = 238; // 255 (NAME_MAX) - 17 (longest suffix: "gcl-" + "-" + jobId + "-build")
+
     static safeDockerString (jobName: string) {
-        return jobName.replace(/[^\w-]+/g, (match) => {
+        const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
+        if (encoded.length <= Utils.MAX_FILENAME_LENGTH) {
+            return encoded;
+        }
+        const hash = createHash("sha256").update(jobName).digest("hex").substring(0, 16);
+        const prefix = encoded.substring(0, Utils.MAX_FILENAME_LENGTH - 1 - hash.length);
+        return `${prefix}-${hash}`;
     }
 
     static safeBashString (s: string) {
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -50,9 +50,13 @@ export class Utils {
         return url.replace(/^https:\/\//g, "").replace(/^http:\/\//g, "");
     }
 
-    static readonly MAX_FILENAME_LENGTH = 238; // 255 (NAME_MAX) - 17 (longest suffix: "gcl-" + "-" + jobId + "-build")
+    // gcl-${safeJobName}-${jobId}-build → wrapper is 17 chars (jobId max 6 digits)
+    static readonly MAX_FILENAME_LENGTH = 255 - 17; // NAME_MAX (bytes) - wrapper
 
     static safeDockerString (jobName: string) {
+        // INVARIANT: \w without /u is ASCII-only ([A-Za-z0-9_]), so `encoded` is pure ASCII
+        // and .length === byte length. NAME_MAX is a byte limit — adding /u would break this.
+        // We hash `jobName` (not `encoded`) because base64url encoding isn't injective.
         const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
--- a/tests/test-cases/parallel-matrix-long-name/.gitlab-ci.yml
+++ b/tests/test-cases/parallel-matrix-long-name/.gitlab-ci.yml
@@ -1,0 +1,10 @@
+---
+build-job:
+  stage: build
+  script:
+    - echo "APP='${APP}' $CI_NODE_INDEX/$CI_NODE_TOTAL"
+  parallel:
+    matrix:
+      - APP:
+          - "my-app-controller,My app controller to be used as reference for development teams,python311,controller,common,controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,controller/coverage/*,controller/build/**/*,controller/coverage/coverage-unit.xml,75,true"
+          - short
--- a/tests/test-cases/parallel-matrix-long-name/.gitlab-ci.yml
+++ b/tests/test-cases/parallel-matrix-long-name/.gitlab-ci.yml
@@ -6,5 +6,7 @@ build-job:
   parallel:
     matrix:
       - APP:
-          - "my-app-controller,My app controller to be used as reference for development teams,python311,controller,common,controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,controller/coverage/*,controller/build/**/*,controller/coverage/coverage-unit.xml,75,true"
+          - "my-app-controller,My app controller to be used as reference for development teams,python311,\
+            controller,common,controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,\
+            controller/coverage/*,controller/build/**/*,controller/coverage/coverage-unit.xml,75,true"
           - short
--- a/tests/utils.test.ts
+++ b/tests/utils.test.ts
@@ -275,7 +275,7 @@ describe("safeDockerString", () => {
 
     it("should encode non-alphanumeric characters", () => {
         const result = Utils.safeDockerString("job/name");
-        expect(result).toContain("Lw"); // '/' encodes to base64url
+        expect(result).toBe("jobLwname"); // '/' → 'Lw'
     });
 
     it("should truncate and hash when encoded name exceeds MAX_FILENAME_LENGTH", () => {
--- a/tests/utils.test.ts
+++ b/tests/utils.test.ts
@@ -266,3 +266,66 @@ describe("getServiceAlias", () => {
         expect(Utils.getServiceAlias({...base, name: "library/nginx", alias: "my-nginx"})).toBe("my-nginx");
     });
 });
+
+describe("safeDockerString", () => {
+    it("should return encoded name unchanged when within limit", () => {
+        const result = Utils.safeDockerString("short-job-name");
+        expect(result).toBe("short-job-name");
+    });
+
+    it("should encode non-alphanumeric characters", () => {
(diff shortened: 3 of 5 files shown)


src/utils.ts
```diff
@@ -49,10 +50,18 @@ export class Utils {
         return url.replace(/^https:\/\//g, "").replace(/^http:\/\//g, "");
     }
 
+    static readonly MAX_FILENAME_LENGTH = 238; // 255 (NAME_MAX) - 17 (longest suffix: "gcl-" + "-" + jobId + "-build")
```
firecow: `238` is hand-counted and the worst-case volume lands at exactly 255 — zero headroom. Derive it from the wrapper:

```suggestion
    // gcl-${safeJobName}-${jobId}-build → wrapper is 17 chars (jobId max 6 digits)
    static readonly MAX_FILENAME_LENGTH = 255 - 17; // NAME_MAX (bytes) - wrapper
```

src/utils.ts
```diff
@@ -49,10 +50,18 @@ export class Utils {
...
-        return jobName.replace(/[^\w-]+/g, (match) => {
+        const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
+        if (encoded.length <= Utils.MAX_FILENAME_LENGTH) {
+            return encoded;
+        }
+        const hash = createHash("sha256").update(jobName).digest("hex").substring(0, 16);
+        const prefix = encoded.substring(0, Utils.MAX_FILENAME_LENGTH - 1 - hash.length);
+        return `${prefix}-${hash}`;
```
firecow, line 68: Load-bearing invariant worth a comment: output is ASCII (`\w` is ASCII-only without `/u`), so `.length` == bytes, which is what `NAME_MAX` limits. Adding `/u` later would break this silently. (Also: hashing `jobName` not `encoded` is correct — `encoded` isn't injective.)

tests/utils.test.ts
```diff
@@ -266,3 +266,66 @@ describe("getServiceAlias", () => {
...
+
+describe("safeDockerString", () => {
+    it("should return encoded name unchanged when within limit", () => {
+        const result = Utils.safeDockerString("short-job-name");
+        expect(result).toBe("short-job-name");
+    });
+
+    it("should encode non-alphanumeric characters", () => {
+        const result = Utils.safeDockerString("job/name");
+        expect(result).toContain("Lw"); // '/' encodes to base64url
```
firecow: `toContain` is brittle; assert the exact transform:

```suggestion
        expect(result).toBe("jobLwname"); // '/' → 'Lw'
```

tests/utils.test.ts
```diff
@@ -266,3 +266,66 @@ describe("getServiceAlias", () => {
         expect(Utils.getServiceAlias({...base, name: "library/nginx", alias: "my-nginx"})).toBe("my-nginx");
     });
 });
+
+describe("safeDockerString", () => {
```
firecow, line 270: Correction — I missed that you already added `integration.test.ts` here covering exactly this (both matrix entries complete, no ENAMETOOLONG). Disregard my earlier ask; this is the right regression guard. 👍

firecow (commented):
Fix is correct, i have a few comments

gyanranjan:
Thanks for the thorough review @firecow — all four points addressed in the latest push:
 
1. Derived constant — 255 - 17 instead of magic 238
2. Invariant comment — documented why \w without /u guarantees .length == bytes, and why we hash jobName not encoded
3. Exact assertion — toBe("jobLwname") instead of toContain("Lw")
4. E2E integration test — added tests/test-cases/parallel-matrix-long-name/ with a matrix value long enough to trigger truncation, asserting the full handler completes without crash
 
  Really appreciate the note about the /u invariant — exactly the kind of subtle correctness detail that would bite someone later. All tests green (61 unit + 33 integration).

tests/test-cases/parallel-matrix-long-name/.gitlab-ci.yml
```diff
@@ -0,0 +1,10 @@
+---
+build-job:
+  stage: build
+  script:
+    - echo "APP='${APP}' $CI_NODE_INDEX/$CI_NODE_TOTAL"
+  parallel:
+    matrix:
+      - APP:
+          - "my-app-controller,My app controller to be used as reference for development teams,python311,controller,common,controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,controller/coverage/*,controller/build/**/*,controller/coverage/coverage-unit.xml,75,true"
```
firecow: yamllint fails here — line too long (299 > 160). A double-quoted scalar with `\`-escaped line continuations keeps the value **byte-identical** (the escaped newline and leading whitespace are both stripped) while getting each line under 160. Verified the parsed value is unchanged and both matrix entries survive.

```suggestion
          - "my-app-controller,My app controller to be used as reference for development teams,python311,\
            controller,common,controller/setup.py,controller/setup_c.py,controller/setup_n.py,controller/tests/**/*,\
            controller/coverage/*,controller/build/**/*,controller/coverage/coverage-unit.xml,75,true"
```

firecow: @gyanranjan I have just comitted this, to see if it resolves the yamllint findings

firecow (approved):

PR #1876, opened 2026-06-05
Improve rulesIf evaluation

ticapix:
Hello,

This is a proposal to improve the robustness of the rules evaluations.

This address https://github.com/firecow/gitlab-ci-local/issues/1859

The main changes are:

- considering more complex rules (eg. from [tbc templates](https://gitlab.com/to-be-continuous/semantic-release/-/blob/4.3.0/templates/gitlab-ci-semrel.yml?ref_type=tags#L130)), add the construction and evaluation of the logical AST based on `&&`, `||`, `(` and `)` operators.
- delegating the parsing into an AST to the small library [jsep](https://github.com/EricSmekens/jsep)
- keep the RE2JS evaluation logic
- keep the regex flags parsing logic
- keep the $VAR expansion logic
- update the tests by:
    - moving `/Hello (?i)world/` as invalid regex, even if RE2JS supports it. (regex101.com, regexr.com are considering it as invalid)
    - since we don't rely on `eval()` only to evaluate the full rule, we can't spy on it. Removing the expected evaluation string. Keep the final expected result verification.

---
## Summary by cubic
Make rules:if evaluation AST-based with `jsep` + `@jsep-plugin/regex` for correct precedence, safer regex handling, clearer errors, and fast failures on unsupported nodes. Keeps RE2JS matching and `$VAR` expansion, and fixes edge cases in complex rules.

- **Bug Fixes**
  - Correctly handle &&, ||, and parentheses via AST instead of regex rewrites.
  - Fix: null !~ /pattern/ returns true; reject `${VAR}` syntax.
  - Enforce RHS of =~/!~ to be a regex; reject quoted strings and invalid flags with helpful errors.
  - Wrap `jsep` parse errors with a consistent “Error attempting to evaluate the following rules:” message.
  - Fail fast on unsupported AST node types in `_nodeToAtom`.
  - Preserve RE2JS regex flag parsing and `$VAR` expansion.
  - Tests assert results only; drop eval-string spying; mark `/Hello (?i)world/` invalid; add “ci-skip” string-quoted regex case.
  - Fix assertion message for non-regex RHS to use the actual RHS value and avoid a ReferenceError crash in =~/!~ evaluations.

- **Dependencies**
  - Add `jsep` and `@jsep-plugin/regex`.

<sup>Written for commit 7f17da96c7f5e2d18158ece3f402fb0fcbe36e4b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecow/gitlab-ci-local/pull/1876?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -291,7 +291,7 @@ export class Utils {
                     });
 
                     const assertMsg = [
-                        "RHS (${rhs}) must be a regex pattern. Do not rely on this behavior!",
+                        `RHS (${rhs}) must be a regex pattern. Do not rely on this behavior!`,
                         "Refer to https://docs.gitlab.com/ee/ci/jobs/job_rules.html#unexpected-behavior-from-regular-expression-matching-with- for more info...",
                     ];
                     assert(_rhs !== regexStr, assertMsg.join("\n"));
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -297,6 +297,12 @@ export class Utils {
                         return `RE2JS.compile(${JSON.stringify(pattern)}, ${flagsBinary})`;
                     });
 
+                    const assertMsg = [
+                        "RHS (${rhs}) must be a regex pattern. Do not rely on this behavior!",
+                        "Refer to https://docs.gitlab.com/ee/ci/jobs/job_rules.html#unexpected-behavior-from-regular-expression-matching-with- for more info...",
+                    ];
+                    assert(_rhs !== regexStr, assertMsg.join("\n"));
+
                     const _operator = n.operator === "=~" ? "!=" : "=="; // =~ -> !=; !~ -> ==
 
                     const evalStr = `${leftStr.raw}.matchRE2JS(${_rhs}) ${_operator} null`;
@@ -328,7 +334,22 @@ export class Utils {
             return res;
         };
 
-        return walk(jsep(evalStr));
+        let ast;
+        try {
+            ast = jsep(evalStr);
+        } catch {
+            const assertMsg = [
+                "Error attempting to evaluate the following rules:",
+                "  rules:",
+                `    - if: '${ruleIf}'`,
+                "as",
+                "```javascript",
+                `${evalStr}`,
+                "```",
+            ];
+            assert(false, assertMsg.join("\n"));
+        }
+        return walk(ast!);
     }
 
     static _evaluateRuleIf (ruleIf: string | undefined, envs: {[key: string]: string}): boolean {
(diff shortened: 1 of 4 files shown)


ticapix:
I'm not sure how I can fix the CI timeout errors :(

src/utils.ts
```diff
@@ -231,97 +266,100 @@ export class Utils {
...
+                    regexStr = this.stripQuotes(regexStr);
+
+                    const regex = /\/(?<pattern>.*)\/(?<flags>[igmsuy]*)/;
+                    const _rhs = regexStr.replace(regex, (_: string, pattern: string, flags: string) => {
+                        const flagsBinary = flagsToBinary(flags);
+                        return `RE2JS.compile(${JSON.stringify(pattern)}, ${flagsBinary})`;
+                    });
+
+                    const assertMsg = [
+                        "RHS (${rhs}) must be a regex pattern. Do not rely on this behavior!",
```
firecow: ${rhs} wont interpolate here

ticapix: Good catch. fixed.

firecow:
@ticapix https://github.com/firecow/gitlab-ci-local/actions/runs/27983948142/job/82891754827?pr=1876

ottok:
I tested this branch in https://github.com/firecow/gitlab-ci-local/issues/1859#issuecomment-5123132198 and can confirm it fixes at least the variant of the regex issue I experienced.

firecow (commented):
AST is right, but two regressions vs master:

**RE2-only regex aborts the rule.** `@jsep-plugin/regex` validates with JS `RegExp`, matching is RE2JS. `/(?i)x/`, `/(?P<n>x)/`, `/a(?s)b/` all rejected. `$CI_COMMIT_TAG =~ /(?i)release/` → `true` on master, throws here. So moving `/Hello (?i)world/` to the invalid set hides it, since regex101 validates JS, not RE2.

**Unary `!` throws.** `!($A == "x")` → `Unsupported expression node type`.

Fix: lex regex literals without JS-validating them; add a `UnaryExpression` case. 

Minor: `eval` still on both `walk` branches; `delete globalThis.RE2JS` wants `finally`.

Also you have conflicts 😄

kilianpaquier:
Hello, confirming this PR fixes https://github.com/firecow/gitlab-ci-local/issues/1859 on current To Be Continuous workflow rules.

@ticapix Any news on your end regarding PR review ?

If needed, rebase with conflicts done and review taken into account here: https://github.com/firecow/gitlab-ci-local/compare/master...kilianpaquier:gitlab-ci-local:fix/rules-if-jsep-evaluator

Issue #350, opened 2021-12-22
Error in substitution when rules contains a regex with a eol-match


d-schiffner:
**Minimal .gitlab-ci.yml illustrating the issue**
```yml
---
job:
  rules:
    - if: $CI_COMMIT_BRANCH =~ /master$/
  script:
    - echo "Heya"
```

**Expected behavior**
The job should be executed if on the master branch

**Host information**
Ubuntu
gitlab-ci-local 4.28.0

**Additional context**
I traced down the issue to the `null` substitution in `utils.ts` in the evaluateRuleIf function: https://github.com/firecow/gitlab-ci-local/blob/master/src/utils.ts#L131

The regex substitution currently uses `\w*` which can be empty, so a single $ will be replaced with null. Changing the expression to `\w+` should fix the issue

firecow:
Fix incoming ⬆️ Thanks for ze report

PR #1889, opened 2026-06-27
fix: safely interpolate variables inside rules regex literals

### LINKED ISSUE #1893: Nested `$variable` inside a `=~` regex pattern produces an invalid regex and crashes rule evaluation
gitlab-ci-local crashes evaluating any rule whose `=~` pattern contains a nested `$variable` reference, for example `/^\$TARGET$/`. This happens regardless of `&&`/`||` or any other operand in the rule. I first suspected a short-circuit bug (the crash showed up inside an `&&` expression), but a minimal repro shows the `&&` is irrelevant, the `=~` match alone is enough.

## Minimal reproduction

`.gitlab-ci.yml`:
```yaml
variables:
  TARGET: '8.3'
  VERSION: '8.3'

test-job:
  stage: test
  script:
    - echo "hello"
  rules:
    - if: $VERSION =~ '/^\$TARGET$/'
      when: never
    - when: on_success
```

Run `gitlab-ci-local` in that directory:

```
Error attempting to evaluate the following rules:
  rules:
    - if: '"8.3" =~ '/^\"8.3"$/''
as
"8.3".matchRE2JS(RE2JS.compile("^\"8.3"$", 0)) != null
```

## Root cause (from the error output)

`$TARGET` gets substituted into the regex source as a quoted string literal (`$TARGET` becomes `"8.3"`), leaving the pattern text as `^\"8.3"$`. The `\"` sequence isn't a valid escape for RE2 (used via re2js), so `RE2JS.compile` throws and the whole pipeline evaluation aborts instead of just failing that one rule.

## Expected behavior

Some resolution of the regex, either a match/no-match result or a rule-level error, without crashing the whole run. Note this exact YAML pattern (nested variable inside a `=~` pattern) is used in production GitLab CI templates today, so real GitLab CI must be handling this substitution differently than gitlab-ci-local does.

## Actual behavior

Hard crash. Pipeline evaluation stops entirely.

## Environment

gitlab-ci-local 4.73.0, Homebrew, macOS arm64

## Real-world impact

This exact pattern is in Drupal.org's shared `gitlab_templates` project, included by every Drupal contrib module's `.gitlab-ci.yml` ([`include.drupalci.main.yml#L495-497`](https://git.drupalcode.org/project/gitlab_templates/-/blob/main/includes/include.drupalci.main.yml#L495-497)):

```yaml
.check-max-php-version-rule: &check-max-php-version-rule
  if: $CORE_PHP_MIN == $CORE_PHP_MAX && ($PHP_VERSION =~ '/^\$_TARGET_PHP$/' || $PHP_VERSION =~ '/^\$CORE_PHP_MAX$/')
  when: never
```

This is a shared anchor evaluated during pipeline parsing, so it blocks gitlab-ci-local on any job in any Drupal contrib pipeline that includes these templates, not just the job the anchor was written for.

Possibly related to the regex-handling issues in #932, though that one doesn't show this exact error.

bbrala:
## `=~` against a regex literal containing a variable crashes the parse

A rule like this aborts the entire pipeline parse:

```yaml
rules:
  - if: $PHP_VERSION =~ '/^$TARGET_PHP$/'
```

With `PHP_VERSION` and `TARGET_PHP` set to `8.3`, gitlab-ci-local builds this JavaScript:

```js
"8.3".matchRE2JS(RE2JS.compile("^\"8.3"$", 0)) != null
```

The pattern is `^\"8.3"$`. Those quotes don't belong there. The injected `"` closes the JS string literal early, `eval` throws, and the whole file fails to load. One rule takes down the run.

### A note on GitLab semantics

GitLab itself does **not** expand variables inside a regex literal — the docs say so plainly: *"Variables in a regular expression are not expanded."* gitlab-ci-local does expand them, and has for a long time (see the `$CI_COMMIT_BRANCH =~ /$BRANCHNAME/` case in #350). This PR doesn't change that decision — it keeps the existing expand-inside-regex behavior and makes it safe. If the project would rather drop the expansion to match GitLab exactly, that's a separate, larger change and a maintainer call; happy to take that direction instead.

### Root cause

`Utils.evaluateRuleIf` in `src/utils.ts` expands every `$VAR` with `JSON.stringify(value)`:

```js
evalStr = this.expandTextWith(evalStr, {
    unescape: JSON.stringify("$"),
    variable: (name) => JSON.stringify(envs[name] ?? null)...,
});
```

That's correct for `==` operands — `8.3 == 8.3` has to become `"8.3" == "8.3"`. But the same global pass also rewrites a `$VAR` sitting inside a `/regex/` literal, and the quotes meant for the string comparison leak into the pattern. The expander can't tell whether a value lands next to a `==` or inside a regex.

### The fix

A pre-pass expands `$VAR` (and `\$VAR`) inside a `/regex/` on the RHS of `=~`/`!~` **before** the general quoting pass, so no variable survives into a literal regex for that pass to quote. The substituted value is **regex-escaped**, so it matches literally and can't break out of the literal or inject anything:

- `/^$TARGET_PHP$/` with `TARGET_PHP=8.3` compiles to `^8\.3$` and matches `8.3`.
- A branch value with a slash — `$CI_COMMIT_BRANCH =~ /^$BRANCH$/`, `BRANCH=feat/x` — no longer closes the literal early. Before this PR it threw; a slash in a branch name is completely ordinary.
- A crafted value like `zzz/ || true || /x` can no longer inject expression syntax into the eval.

The other shapes are untouched:

1. Plain `==` operands still get quoted.
2. `=~` with a literal `/regex/` and no variable is unchanged.
3. `=~` where the RHS is a variable that holds a regex (`$TAG =~ $TAG_REGEX`) still routes through the existing path.

### Tests

Added cases to `tests/rules-regex.test.ts`: a variable interpolated into the pattern, a value containing `/` (matches literally, no throw), an injection attempt that must stay `false`, literal metacharacter matching, and `!~` negation with a slash value.

One existing assertion changed. The #350 case (`$CI_COMMIT_BRANCH =~ /$BRANCHNAME/`) asserted `when: "never"` — that froze the old buggy behavior. With `BRANCHNAME=master` matching `master`, it's `when: "manual"`; the test asserts that now, with a comment.

### Known limitation (not addressed here)

A backslash inside a value is still mishandled on the **left** side of the match — `JSON.stringify(value).replaceAll("\\\\", "\\")` turns `a\b` into the JS escape `a\b` (a backspace), so a literal self-match returns `false`. That's a pre-existing quirk in the general LHS expansion, unrelated to regex interpolation, and changing it touches every backslash-bearing value — out of scope for this fix.

`tsc --noEmit` and `eslint` are clean; the rules suites pass (71 tests).

Fixes #1893

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -267,7 +267,11 @@ export class Utils {
             const expandedPattern = pattern.replaceAll(
                 /(\$\$)|\\?\$\{([a-zA-Z_]\w*)}|\\?\$([a-zA-Z_]\w*)/g,
                 (_m: string, escape: string, var1: string, var2: string) => {
-                    if (escape !== undefined) return "$";
+                    // Leave a `$$` escape intact for the global unescape pass
+                    // (`$$` -> literal `$`). Returning a single `$` here would
+                    // re-expose the following name as a `$VAR` to that pass and
+                    // expand it, defeating the escape.
+                    if (escape !== undefined) return "$$";
                     return escapeRegExp(envs[var1 || var2] ?? "");
                 },
             );
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -245,6 +245,30 @@ export class Utils {
             return binary;
         };
 
+        // A `$VAR` that sits *inside* a regex literal on the RHS of `=~`/`!~`
+        // must be substituted with its raw value, not a quoted JS string
+        // literal. The general expansion below wraps every value in quotes
+        // (needed for `==` operands), which would otherwise leak `"`
+        // characters into the regex pattern, e.g. turning
+        //   $PHP_VERSION =~ /^\$_TARGET_PHP$/   (with _TARGET_PHP=8.3)
+        // into the malformed pattern `^\"8.3"$` instead of `^8.3$`.
+        // In a regex context GitLab still expands an escaped `\$VAR`, dropping
+        // the escaping backslash. We do this as a pre-pass, before the general
+        // expansion, so that no `$VAR` remains inside literal regexes; this
+        // leaves the "RHS is a variable that holds a regex" case (e.g.
+        // `$TAG =~ $TAG_REGEX`) to the general quoted expansion as before.
+        const regexLiteralRhs = /(?<op>=~|!~)(?<pre>\s*["']?)\/(?<pattern>(?:\\.|[^/\\])*)\//g;
+        evalStr = evalStr.replaceAll(regexLiteralRhs, (_match, op, pre, pattern) => {
+            const expandedPattern = pattern.replaceAll(
+                /(\$\$)|\\?\$\{([a-zA-Z_]\w*)}|\\?\$([a-zA-Z_]\w*)/g,
+                (_m: string, escape: string, var1: string, var2: string) => {
+                    if (escape !== undefined) return "$";
+                    return envs[var1 || var2] ?? "";
+                },
+            );
+            return `${op}${pre}/${expandedPattern}/`;
+        });
+
         // Expand all variables
         evalStr = this.expandTextWith(evalStr, {
             unescape: JSON.stringify("$"),
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -246,10 +246,10 @@ export class Utils {
         };
 
         // A `$VAR` that sits *inside* a regex literal on the RHS of `=~`/`!~`
-        // must be substituted with its raw value, not a quoted JS string
-        // literal. The general expansion below wraps every value in quotes
-        // (needed for `==` operands), which would otherwise leak `"`
-        // characters into the regex pattern, e.g. turning
+        // must be substituted with its value, not a quoted JS string literal.
+        // The general expansion below wraps every value in quotes (needed for
+        // `==` operands), which would otherwise leak `"` characters into the
+        // regex pattern, e.g. turning
         //   $PHP_VERSION =~ /^\$_TARGET_PHP$/   (with _TARGET_PHP=8.3)
         // into the malformed pattern `^\"8.3"$` instead of `^8.3$`.
         // In a regex context GitLab still expands an escaped `\$VAR`, dropping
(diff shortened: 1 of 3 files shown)


bbrala:
Good catch by cubic — confirmed and fixed in 71dddce.

The bug was real: the pre-pass rewrote `$$` to a single `$`, which the later global expansion pass then read as `$VAR` and expanded, defeating the escape. That was a regression versus the previous single-pass behavior.

The suggested remedy (dropping `(\$\$)|` from the pre-pass regex) doesn't fully work, though: with that branch gone, the pre-pass's variable alternative `\$([a-zA-Z_]\w*)` still matches the **second** `$` of `$$FOO` as a fresh `$FOO` and expands it, leaving a stray `$` — so `$$FOO` becomes `$<value>` rather than a literal.

Instead I kept the `$$` branch but return `"$$"` (rather than `"$"`), so it consumes both dollars as a unit — the inner matcher can't grab the second one — and hands the escape to the global unescape pass exactly as before the pre-pass existed. The following name is no longer expanded. Added a regression test for it.

src/utils.ts
```diff
@@ -245,6 +245,39 @@ export class Utils {
...
+                (_m: string, escape: string, var1: string, var2: string) => {
+                    // Leave a `$$` escape intact for the global unescape pass
+                    // (`$$` -> literal `$`). Returning a single `$` here would
+                    // re-expose the following name as a `$VAR` to that pass and
+                    // expand it, defeating the escape.
+                    if (escape !== undefined) return "$$";
+                    return escapeRegExp(envs[var1 || var2] ?? "");
+                },
+            );
+            return `${op}${pre}/${expandedPattern}/`;
```
firecow, line 278: Unset var → `//` → reaches `eval` as `"anything" =~ //` and aborts the run. Master returns `false`.

```suggestion
            return `${op}${pre}/${expandedPattern || "(?:)"}/`;
```

Still `false` (`matchRE2JS` skips zero-length matches, `global.ts:21`).

src/utils.ts
```diff
@@ -245,6 +245,39 @@ export class Utils {
...
+        // leaves the "RHS is a variable that holds a regex" case (e.g.
+        // `$TAG =~ $TAG_REGEX`) to the general quoted expansion as before.
+        // The substituted value is regex-escaped so it matches literally and a
+        // value containing regex metacharacters (a branch like `feat/x`, or a
+        // value carrying `/`, `|`, `(`, ...) cannot break out of the literal,
+        // close it early, or inject expression syntax into the eval.
+        const escapeRegExp = (value: string): string => value.replace(/[.*+?^${}()|[\]\\/]/g, "\\$&");
+        const regexLiteralRhs = /(?<op>=~|!~)(?<pre>\s*["']?)\/(?<pattern>(?:\\.|[^/\\])*)\//g;
+        evalStr = evalStr.replaceAll(regexLiteralRhs, (_match, op, pre, pattern) => {
+            const expandedPattern = pattern.replaceAll(
```
firecow, line 267: Escaping breaks regex fragments: `/^($ALLOWED)$/` with `a|b` is `false` here, `true` on GitLab and inconsistent with `$TAG =~ $TAG_REGEX`, which stays unescaped.

src/utils.ts
```diff
@@ -245,6 +245,39 @@ export class Utils {
             return binary;
         };
 
+        // A `$VAR` that sits *inside* a regex literal on the RHS of `=~`/`!~`
```
firecow, line 248: Repo doesn't carry source comments

firecow (commented):
Fixes a real bug: every `$VAR` inside a regex literal returns `false` on master. One blocker inline, plus two notes.

PR #1905, opened 2026-07-31
fix: system-wide deadlock

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `High` | **File**: `src/mutex.ts:L20`

The `Mutex.exclusive` method adds a key to the locks set and then executes the callback `cb()`. If `cb()` throws an exception, the `Mutex.locks.delete(key)` statement is never reached. This leaves the lock permanently acquired, causing any future calls to `waitForLock` for the same key to poll indefinitely, resulting in a system-wide deadlock for that key.

## Solution

Wrap the callback execution in a `try...finally` block to ensure the lock is always released, even if the callback throws an error.
```typescript
static async exclusive (key: string, cb: () => Promise<void>) {
    await Mutex.waitForLock(key);
    Mutex.locks.add(key);
    try {
        await cb();
    } finally {
        Mutex.locks.delete(key);
    }
}
```

## Changes

- `src/mutex.ts` (modified)
- `src/state.ts` (modified)

---
## Summary by cubic
Fixes two security issues: ensures mutex locks are always released even if callbacks throw, and makes YAML state loading safe with a default for empty files.

- **Bug Fixes**
  - Wrap `Mutex.exclusive` callback in try/finally to always release the lock.
  - Load state with `yaml.load(content) || {}` for safe parsing and to avoid null/unsafe types on empty files.

<sup>Written for commit 69ee1462f90b4f7fc044962f99ca67510f690eeb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecow/gitlab-ci-local/pull/1905?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


src/state.ts
```diff
@@ -7,7 +7,7 @@ const loadStateYML = async (stateFile: string): Promise<any> => {
         return {};
     }
     const stateFileContent = await fs.readFile(stateFile, "utf8");
-    return yaml.load(stateFileContent) || {};
+    return yaml.load(stateFileContent, {schema: yaml.DEFAULT_SAFE_SCHEMA}) || {};
```
firecow:
```suggestion
    return yaml.load(stateFileContent) || {};
```

firecow (commented):
Mutex fix is good. Revert the state.ts hunk: `load()` is already safe by default in js-yaml v4.

src/state.ts
```diff
@@ -34,4 +34,4 @@ const incrementPipelineIid = async (cwd: string, stateDir: string): Promise<numb
     });
 };
 
-export {getPipelineIid, incrementPipelineIid};
+export {getPipelineIid, incrementPipelineIid};
```
firecow:
```suggestion
export {getPipelineIid, incrementPipelineIid};

```

firecow (approved):

PR #1928, opened 2026-08-23
fix: write the rsync exclude list to a file

### LINKED ISSUE #1925: Fail with rsync 3.5.0
Hello

(This is mostly to let everyone knows)

Due to https://github.com/RsyncProject/rsync/issues/1053 gitlab-ci-local fails with:
```
e2: Command failed with exit code 11: 'rsync -a --delete-excluded --delete --exclude-from=<(git ls-files -o --directory | awk '\''{print "/"$0}'\'') --exclude .git/lfs --exclude .gitlab-ci-local/ ./ .gitlab-ci-local/builds/.docker/'

rsync: [client] failed to open exclude file /dev/fd/63: No such file or directory (2)
rsync error: error in file IO (code 11) at exclude.c(1719) [client=3.5.0-g471e17dc]
    at IU (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:206:2356)
    at Z4 (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:230:1371)
    at Te (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:242:42142)
    at Ie (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:242:41862)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async R.rsyncTrackedFiles (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:391:927)
    at async CZ (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:328:1087)
    at async Object.handler (file:///usr/lib/node_modules/gitlab-ci-local/dist/index.js:396:3699)
```

current workaround (that I know of): downgrade rsync.

firecow:
rsync 3.5.0 refuses to open an exclude file passed as a process substitution (`/dev/fd/63`), so every sync failed with exit code 11; the untracked list is now written into the state dir and passed as a real path. Fixes #1925.

---
## Summary by cubic
Fixes rsync sync failures by writing the untracked exclude list to a real file and passing its path to `rsync`. Previously we used process substitution (`/dev/fd/*`), which `rsync` 3.5.0 refuses to read, causing exit code 11.

- Generate untracked paths via `git ls-files -o --directory`, prefix with "/", and write to `${stateDir}/rsync-exclude-${target}` (file is now named directly after the target; extra sanitizing removed).
- Pass `--exclude-from=<that-file>` to `rsync`; still honor the optional `ignoresFile`.
- Add an integration test that asserts tracked files sync, untracked files are excluded, and the exclude file exists with the expected entry.
- Side effect: exclude files persist under the state directory; no config or CLI changes required.

<sup>Written for commit 3c2ba4658f2e6a1ed181cf9b06068b7526327d09. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecow/gitlab-ci-local/pull/1928?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -382,7 +382,7 @@ export class Utils {
         const time = process.hrtime();
         await fs.mkdirp(`${cwd}/${stateDir}/builds/${target}`);
         const {stdout: untracked} = await Utils.bash("git ls-files -o --directory", cwd);
-        const untrackedFile = `${cwd}/${stateDir}/rsync-exclude-${target.replaceAll(/[^\w.-]/g, "_")}`;
+        const untrackedFile = `${cwd}/${stateDir}/rsync-exclude-${target}`;
         await fs.writeFile(untrackedFile, untracked.split("\n").filter(line => line !== "").map(line => `/${line}`).join("\n"));
         const cmd = [
             "rsync -a --delete-excluded --delete",
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -381,9 +381,12 @@ export class Utils {
     static async rsyncTrackedFiles (cwd: string, stateDir: string, ignoresFile: string, target: string): Promise<{hrdeltatime: [number, number]}> {
         const time = process.hrtime();
         await fs.mkdirp(`${cwd}/${stateDir}/builds/${target}`);
+        const {stdout: untracked} = await Utils.bash("git ls-files -o --directory", cwd);
+        const untrackedFile = `${cwd}/${stateDir}/rsync-exclude-${target.replaceAll(/[^\w.-]/g, "_")}`;
+        await fs.writeFile(untrackedFile, untracked.split("\n").filter(line => line !== "").map(line => `/${line}`).join("\n"));
         const cmd = [
             "rsync -a --delete-excluded --delete",
-            "--exclude-from=<(git ls-files -o --directory | awk '{print \"/\"$0}')",
+            `--exclude-from=${Utils.safeBashString(untrackedFile)}`,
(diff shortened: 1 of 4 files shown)


firecow:
Context on why this is not worth waiting out upstream.

rsync 3.5.0 fixed CVE-2026-53802 (HIGH), which covered exactly this flag:

> Arbitrary file read / transfer-shaping via symlinked operator-supplied input files. rsync followed attacker-planted symlinks in `--filter` merge files, `--files-from` / `--include-from` / `--exclude-from`, and the client `--password-file` / daemon secrets file [...] Operator-supplied paths are now resolved component-by-component with `openat(O_PATH|O_NOFOLLOW)`, allowing a symlink component only when it is owned by uid 0 or the effective uid.

That resolver is the new `ona_open()`. Per RsyncProject/rsync#1053, `readlinkat()` on `/dev/fd/63` returns the kernel pseudo-path `pipe:[NNNNNN]`, which is then treated as a literal relative filename and fails with ENOENT. The issue is still open, so a targeted fix for the pipe case is needed; the hardening itself is not going away.

Also worth knowing, since the same check applies to any path we pass: on 3.5.0 a project reached through a symlink owned by another user will now be refused too. That is not new here, the `--exclude-from` for the ignores file has the same exposure, but it is a new class of exit code 11.

PR #1929, opened 2026-08-23
fix: stop encoded job names colliding with literal ones

firecow:
base64url output reuses the same characters that pass through unencoded, so a job named `aLgb` took the same safe name as one named `a.b` and the two silently shared a build directory; encoded names now carry a short digest of the original. Job names that need no encoding are unchanged, but any name containing a special character gets a new safe name, so existing build dirs, containers and cache volumes for those jobs are orphaned on first run after upgrade.

---
## Summary by cubic
Stop encoded job names colliding with literal ones in `Utils.safeDockerString`. Previously, `a.b` and `aLgb` could map to the same safe name; now any encoded name is suffixed with `.XXXXXXXX` (8-hex digest), and all hashes use a dot separator so literal names (which never include dots) cannot collide. Names that need no encoding are unchanged. If the result exceeds the filename limit, we truncate the encoded prefix and append `.YYYYYYYYYYYYYYYY` (16-hex digest).

- Rollout and migration
  - Jobs with special characters will get new safe names. Prune orphaned build directories, containers, and cache volumes after deploy.
  - Verify long names still respect the filename limit; truncation now yields `<prefix>.<16-hex>`.

- Review notes
  - Tests assert explicit safe names; artifact paths now include a dot and digest (e.g., `produceIAartifactsIPCfk50.c2393d2b`).

<sup>Written for commit 48e79a84790434fe0e7a8f668dd46c3bc8c8d552. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/firecow/gitlab-ci-local/pull/1929?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -60,9 +60,7 @@ export class Utils {
         const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
-        const unambiguous = encoded === jobName ?
-            encoded :
-            `${encoded}.${createHash("sha256").update(jobName).digest("hex").substring(0, 8)}`;
+        const unambiguous = encoded === jobName ? encoded : ${encoded}.${createHash("sha256").update(jobName).digest("hex").substring(0, 8)}`;
         if (unambiguous.length <= Utils.MAX_FILENAME_LENGTH) {
             return unambiguous;
         }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -60,11 +60,14 @@ export class Utils {
         const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
-        if (encoded.length <= Utils.MAX_FILENAME_LENGTH) {
-            return encoded;
+        const unambiguous = encoded === jobName ?
+            encoded :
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -60,12 +60,15 @@ export class Utils {
         const encoded = jobName.replace(/[^\w-]+/g, (match) => {
             return base64url.encode(match);
         });
-        if (encoded.length <= Utils.MAX_FILENAME_LENGTH) {
-            return encoded;
+        const unambiguous = encoded === jobName ?
+            encoded :
+            `${encoded}.${createHash("sha256").update(jobName).digest("hex").substring(0, 8)}`;
```
firecow:
```suggestion
        const unambiguous = encoded === jobName ? encoded : ${encoded}.${createHash("sha256").update(jobName).digest("hex").substring(0, 8)}`;
```

